PR #11, opened 2022-11-27
Unsupported Curl Option

joshedney:
## 👩 Platform release notes

Update the install script so that curl command doesn't fail when using older versions of curl.

## 🔬 Testing

Covered by CI/ Manual testing

## 🔒 Security
No impact on security

## 🚀 Deployment
Manual release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
 # 1.0.0 TBD
 
+## Fixes
+- Remove unsupported curl options [11](https://github.com/bugsnag/bugsnag-cli/pull/11)
+- Fix using port within endpoint url CLI option [10](https://github.com/bugsnag/bugsnag-cli/pull/10)
 ## Enhancements 
 - Add dart symbol file support [6](https://github.com/bugsnag/bugsnag-cli/pull/6)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,4 @@
 # 1.0.0 TBD
 
-## Fixes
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -136,13 +136,17 @@ ohai "Downloading and installing Bugsnag CLI..."
...
-  DOWNLOAD_URL=$(curl -sS --no-progress-meter ${BUGSNAG_CLI_GIT_REMOTE} |
+  DOWNLOAD_URL=$(curl -sS ${BUGSNAG_CLI_GIT_REMOTE} |
     grep "${UNAME_MACHINE}-${OS_NAME}-bugsnag-cli*" |
     grep "browser_download_url" |
     cut -d : -f 2,3 |
     tr -d \" |
     xargs)
 
+  if [[ ! -n ${DOWNLOAD_URL} ]]; then
+    abort "Failed to get download url"
```
tomlongridge, line 147: I think we could provide more context in the error message here:

```suggestion
    abort "Failed to get download URL from ${BUGSNAG_CLI_GIT_REMOTE}"
```

CHANGELOG.md
```diff
@@ -1,4 +1,7 @@
 # 1.0.0 TBD
 
+## Fixes
+- Remove unsupported curl options [11](https://github.com/bugsnag/bugsnag-cli/pull/11)
+- Fix using port within endpoint url CLI option [10](https://github.com/bugsnag/bugsnag-cli/pull/10)
```
tomlongridge, line 5: The tool isn't released yet, so we don't need to start tracking changes right now. When we tag the first release, we just need an initial entry that summarises the initial functionality.

tomlongridge (approved):

PR #14, opened 2022-12-01
[PLAT-9287] Add build tool command structure

joshedney:
## Goal

Add the command structure for the create build command options.

## Changeset

Add package and command structure for the build tool.

## Testing

Covered by CI

### DIFF
--- a/pkg/build/create.go
+++ b/pkg/build/create.go
@@ -1,13 +1,10 @@
 package build
 
 type CreateBuild struct {
-	AppVersion       string            `help:"The version number of the application."`
-	AppVersionCode   string            `help:"The version code for the application (Android only)."`
-	AppBundleVersion string            `help:"The bundle version for the application (iOS only)."`
 	BuilderName      string            `help:"The name of the entity that triggered the build. Could be a user, system etc."`
-	Metadata         map[string]string `help:"Additional build information"`
-	ReleaseStage     string            `help:"The release stage (eg, production, staging) that is being released (if applicable)."`
-	Provider         string            `help:"he name of the source control provider that contains the source code for the build."`
+	Metadata         map[string]string `help:"(optional) Additional build information"`
+	ReleaseStage     string            `help:"(optional) The release stage (eg, production, staging) that is being released (if applicable)."`
+	Provider         string            `help:"(optional) The name of the source control provider that contains the source code for the build."`
 	Repository       string            `help:"The URL of the repository containing the source code being deployed."`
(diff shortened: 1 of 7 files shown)


test/utils/endpoint_test.go
```diff
@@ -22,7 +22,7 @@ func TestEndpointBuilding(t *testing.T) {
 	if err != nil {
 		t.Error(err)
 	}
-	assert.Equal(t, results, "https://upload.bugsnag.com" + ":8443", "They should be the same")
+	assert.Equal(t, results, "https://upload.bugsnag.com"+":8443", "They should be the same")
```
kattrali, line 25:
```suggestion
	assert.Equal(t, results, "https://upload.bugsnag.com:8443", "They should be the same")
```

kattrali (approved):

pkg/build/create.go
```diff
@@ -0,0 +1,13 @@
...
+type CreateBuild struct {
+	AppVersion       string            `help:"The version number of the application."`
+	AppVersionCode   string            `help:"The version code for the application (Android only)."`
+	AppBundleVersion string            `help:"The bundle version for the application (iOS only)."`
+	BuilderName      string            `help:"The name of the entity that triggered the build. Could be a user, system etc."`
+	Metadata         map[string]string `help:"Additional build information"`
+	ReleaseStage     string            `help:"The release stage (eg, production, staging) that is being released (if applicable)."`
+	Provider         string            `help:"he name of the source control provider that contains the source code for the build."`
+	Repository       string            `help:"The URL of the repository containing the source code being deployed."`
+	Revision         string            `help:"The source control SHA-1 hash for the code that has been built (short or long hash)"`
```
tomlongridge, line 12: Would be good to make these strings consistent with the `DartSymbol` descriptions - at least manually (adding the "optional" prefix and making sure the common options are the same text).

Ideally, perhaps we could extract them into a common structure so they can be shared? I imagine as the tool grows we're going to be repeating options quite often.

pkg/build/create.go
```diff
@@ -0,0 +1,10 @@
+package build
+
+type CreateBuild struct {
+	BuilderName      string            `help:"The name of the entity that triggered the build. Could be a user, system etc."`
+	Metadata         map[string]string `help:"(optional) Additional build information"`
```
tomlongridge, line 5: Are any of the flags _not_ optional? I'm not sure I see the distinction between those marked "(optional)" and those not. Perhaps we should just not show that in the help?

tomlongridge (approved):

PR #15, opened 2022-12-06
Add build command to the CLI

joshedney:
## Goal

Add functionality to the CLI to allow the creation/registration of builds via the bugsnag API

Example:
`bugsnag-cli create-build --api-key=0123456789 --app-version=1.2.3`

Output:
```
[INFO] Creating build on: https://build.bugsnag.com:443
[INFO] Build information:
{
    "apiKey": "0123456789",
    "builderName": "josh.edney",
    "sourceControl": {
        "repository": "https://github.com/bugsnag/bugsnag-cli",
        "revision": "b3242b99ecebf7077e665626f705aef7e120a5a5"
    },
    "appVersion": "1.2.3"
}
[SUCCESS] Build created
```

## Changeset

Added function to get commit has from the system using git cli
Added function to get repo URL from the system using git cli
Added function to get system user using `whoami`
Added function to create and send build JSON payload to bugsnag API

## Testing

Covered by CI


pkg/utils/get-repository-info.go
```diff
@@ -0,0 +1,57 @@
...
+//ParseGitUrl - Parses a git url to ensure that it is in HTTPS format
+func ParseGitUrl(url string) string {
+	if strings.Contains(url, "git@") {
+		url = strings.Replace(url, ":", "/", 1)
+		url = strings.Replace(url, "git@", "https://", 1)
+		url = strings.Replace(url, ".git", "", 1)
+		return url
+	}
+	return url
+}
```
kattrali, line 39: I don't think any of this is necessary (and might be github-specific, breaking other providers)

pkg/utils/get-system-user.go
```diff
@@ -0,0 +1,14 @@
+package utils
+
+import (
+	"os/user"
+)
+
+// GetSystemUser - Gets username from the system
+func GetSystemUser() (string, error) {
+	user, err := user.Current()
```
kattrali, line 9: This is so much more convenient than having to call `whoami` :joy:

test/build/create_build_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+func TestGettingRepoInfo(t *testing.T) {
+	t.Log("Test getting repo info map")
+	results, err := build.GetRepoInfo("", "", "")
+	if err != nil {
+		t.Error(err)
+	}
+	assert.Equal(t, map[string]string{
+		"repository": "https://github.com/bugsnag/bugsnag-cli",
+		"revision":   os.Getenv("GIT_SHA"),
```
kattrali, line 35: This is more general than this one test/assertion, but I'd instead avoid asserting on the revision auto-detected results in unit tests and rely on the features for this. As written, running `go test` directly results in broken tests since the makefile magic is not used.

pkg/build/create.go
```diff
@@ -8,3 +21,128 @@ type CreateBuild struct {
...
+	res, err := server.SendRequest(req, 300)
+
+	if err != nil {
+		return fmt.Errorf("error sending file request: %w", err)
+	}
+
+	b, err := io.ReadAll(res.Body)
+
+	if strings.Contains(string(b), "Source control provider is missing") {
+		log.Warn("Source control provider is missing and could not be inferred. Please resend using one of: [github-enterprise, github, gitlab-onpremise, gitlab, bitbucket-server, bitbucket]. Request was still processed but source control information was ignored.")
```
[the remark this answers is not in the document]
joshedney: I've lowered this to a INFO so that they're still informed like they would be via a standard curl request

pkg/utils/get-repository-info.go
```diff
@@ -0,0 +1,57 @@
...
+
+// GetRepoUrl - Gets the URl of a git repo.
+func GetRepoUrl() (string, error) {
+	gitLocation, err := exec.LookPath("git")
+
+	if err != nil {
+		return "", fmt.Errorf("unable to find git on system: %w", err)
+	}
+
+	cmd := exec.Command(gitLocation, "config", "--get", "remote.origin.url")
```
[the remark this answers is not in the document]
joshedney: I've reworked this  so that if it fails it will then get a list of the remotes and use the first one to get the remote url from. If it fails at any point it will just return an empty string and not error.

kattrali (approved):

PR #25, opened 2023-04-12
[PLAT-9951] Update Android AAB variant handling

joshedney:
## Goal

Adjust the functionality of the Android AAB processing to only work with a single variant, from the root of the Android project folder and pull the API key from the `AndroidManifest.xml` if not provided

Example: 
```bash
# Run from the root of the Android project
$ bugsnag-cli upload android-AAB 
```

Output:
```bash
[INFO] Uploading files to: https://upload.bugsnag.com:443
[INFO] Processing app-release.aab
[INFO] Expanding app-release.aab
[SUCCESS] app-release.aab expanded
[INFO] Processing mapping.txt for variant: release
[INFO] Compressing app/build/outputs/mapping/release/mapping.txt
[INFO] Uploading debug information for app/build/outputs/mapping/release/mapping.txt
[SUCCESS] app/build/outputs/mapping/release/mapping.txt uploaded
[INFO] Using Android NDK located here: /opt/homebrew/share/android-commandlinetools/ndk/25.1.8937393
[INFO] Locating objcopy within Android NDK path
[INFO] Using objcopy located: /opt/homebrew/share/android-commandlinetools/ndk/25.1.8937393/toolchains/llvm/prebuilt/darwin-x86_64/bin/llvm-objcopy
[INFO] Building file list for configuration: release
[INFO] Searching directory for files
[INFO] Processing NDK files for configuration: release
[INFO] Extracting debug info from libentrypoint.so.sym using objcopy
[INFO] app/build/outputs/bundle/release/raw/BUNDLE-METADATA/com.android.tools.build.debugsymbols/arm64-v8a/libentrypoint.so.sym.so
[INFO] Uploading debug information for libentrypoint.so.sym
[SUCCESS] libentrypoint.so.sym uploaded
[INFO] Extracting debug info from libentrypoint.so.sym using objcopy
[INFO] app/build/outputs/bundle/release/raw/BUNDLE-METADATA/com.android.tools.build.debugsymbols/armeabi-v7a/libentrypoint.so.sym.so
[INFO] Uploading debug information for libentrypoint.so.sym
[SUCCESS] libentrypoint.so.sym uploaded
[INFO] Extracting debug info from libentrypoint.so.sym using objcopy
[INFO] app/build/outputs/bundle/release/raw/BUNDLE-METADATA/com.android.tools.build.debugsymbols/x86/libentrypoint.so.sym.so
[INFO] Uploading debug information for libentrypoint.so.sym
[SUCCESS] libentrypoint.so.sym uploaded
[INFO] Extracting debug info from libentrypoint.so.sym using objcopy
[INFO] app/build/outputs/bundle/release/raw/BUNDLE-METADATA/com.android.tools.build.debugsymbols/x86_64/libentrypoint.so.sym.so
[INFO] Uploading debug information for libentrypoint.so.sym
[SUCCESS] libentrypoint.so.sym uploaded
[SUCCESS] Upload(s) completed
```

## Changeset

Adjust how multiple variants are processed to fail if multiple variants are detected.
Pull the API key from the `AndroidManifest.xml` if it is not provided via a flag.
Se the default paths based from the root of a default android project.

## Testing

Covered by CI


pkg/upload/android-aab.go
```diff
@@ -2,138 +2,103 @@ package upload
...
-		}
+	fileList, err := utils.BuildFileList([]string{soFilePath})
 
-		log.Info("Using Android NDK located here: " + androidNdkRoot)
+	if err != nil {
+		return fmt.Errorf("error building `.so` file list. " + err.Error())
+	}
 
-		log.Info("Locating objcopy within Android NDK path")
+	mappingFilePath := filepath.Join(tempDir, "BUNDLE-METADATA", "com.android.tools.build.obfuscation", "proguard.map")
```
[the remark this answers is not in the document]
joshedney: That makes sense I was under the assumption that it would be in the AAB but it doesn't hurt with an additional check

pkg/upload/android-aab.go
```diff
@@ -2,138 +2,136 @@ package upload
...
-
-		if err != nil {
-			return err
+			if err != nil {
+				return err
+			}
 		}
+	} else {
+		log.Info("Skipping NDK processing")
+	}
```
tomlongridge, line 123: If you have no .so files in your AAB (presumably as you have no native code), what does the directory structure look like? Do you have no `com.android.tools.build.debugsymbols` or is the directory empty?

We want to make sure we don't error in the case of there being no NDK files.

I think the Info message is good, but maybe word it in a way that makes it obvious that it's not a problem:
```suggestion
	} else {
		log.Info("No NDK (.so) files detected for upload.")
	}
```

pkg/upload/android-aab.go
```diff
@@ -2,138 +2,136 @@ package upload
...
-		err = android.ProcessNdk(apiKey, variant, outputPath, aabManifestData, objCopyPath, projectRoot, overwrite, timeout, endpoint, failOnUploadError)
+	if utils.FileExists(mappingFilePath) {
+		err = ProcessAndroidProguard(apiKey, applicationId, "", buildUuid, []string{mappingFilePath}, "", versionCode, versionName, endpoint, retries, timeout, overwrite, dryRun)
 
 		if err != nil {
-			return fmt.Errorf("error processing NDK symbol files. " + err.Error())
+			return err
 		}
+	} else {
+		log.Info("Skipping Proguard processing")
```
tomlongridge, line 134:
```suggestion
		log.Info("No Proguard (mapping.txt) file detected for upload.")
```

pkg/upload/android-ndk.go
```diff
@@ -24,9 +25,18 @@ type AndroidNdkMapping struct {
 func ProcessAndroidNDK(apiKey string, applicationId string, androidNdkRoot string, appManifestPath string, paths []string, projectRoot string, variant string, versionCode string, versionName string, endpoint string, failOnUploadError bool, retries int, timeout int, overwrite bool, dryRun bool) error {
 
 	var fileList []string
+	var processedFileList []string
 	var mergeNativeLibPath string
 	var err error
 
+	tempDir, err := os.MkdirTemp("", "bugsnag-cli-ndk-*")
```
tomlongridge, line 32: I'd prefer to see this done closer to the code that uses it – a lot can happen before the tmp dir is needed and potentially it isn't needed at all if .sym files have been provided.

Suggest adding creation of the directory to the file-loop - if the variable is nil when you need to do an objcopy, then create it first:

```
tempDir = nil
		for _, file := range fileList {
			if strings.HasSuffix(file, ".so.sym") {
				processedFileList = append(processedFileList, file)
			} else if filepath.Ext(file) == ".so" {

				log.Info("Extracting debug info from " + filepath.Base(file) + " using objcopy")

// if tempDir == nil
//   tempDir = MkdirTemp("", "bugsnag-cli-ndk-*")

				outputFile, err := android.Objcopy(objCopyPath, file, tempDir)
...
```

Nit - I'd also not call it tempDir, but workingDir or similar - it describes it's purpose rather than what it is (perhaps it might change in future to a user configurable dir).

pkg/upload/android-ndk.go
```diff
@@ -24,9 +25,18 @@ type AndroidNdkMapping struct {
 func ProcessAndroidNDK(apiKey string, applicationId string, androidNdkRoot string, appManifestPath string, paths []string, projectRoot string, variant string, versionCode string, versionName string, endpoint string, failOnUploadError bool, retries int, timeout int, overwrite bool, dryRun bool) error {
 
 	var fileList []string
+	var processedFileList []string
```
tomlongridge, line 28:
```suggestion
	var symbolFileList []string
```

pkg/upload/android-aab.go
```diff
@@ -2,138 +2,136 @@ package upload
...
 
-		outputPath := filepath.Join(strings.Replace(config["aabPath"], filepath.Base(config["aabPath"]), "", -1), "raw")
-
-		log.Info("Expanding " + filepath.Base(config["aabPath"]))
-
-		err := utils.Unzip(config["aabPath"], outputPath)
-
 		if err != nil {
 			return err
 		}
```
tomlongridge, line 77: Can err be non-nil here? Think it would already have returned.

tomlongridge (approved):

lemnik (approved):

PR #27, opened 2023-04-25
Remove `--android-ndk-root` from Android AAB

joshedney:
## Goal

Remove the CLI option `--android-ndk-root` from the Android AAB upload section as it's not required due to the symbol files not needing to be passed through `objcopy`.

During testing, I also noticed that we're missing an `if err !=nil ` for checking if we need to fall back during the Android Proguard processing.

## Changeset

Remove `--android-ndk-root` from CLI options for Android AAB
Check if we have an `err` before we attempt to fall back to the default endpoint for Android Proguard

## Testing

Covered By CI

### DIFF
--- a/pkg/upload/android-aab.go
+++ b/pkg/upload/android-aab.go
@@ -10,16 +10,15 @@ import (
 )
 
 type AndroidAabMapping struct {
-	ApplicationId  string            `help:"Module application identifier"`
-	AndroidNdkRoot string            `help:"Path to Android NDK installation ($ANDROID_NDK_ROOT)"`
-	BuildUuid      string            `help:"Module Build UUID"`
-	Path           utils.UploadPaths `arg:"" name:"path" help:"(required) Path to directory or file to upload" type:"path"`
-	ProjectRoot    string            `help:"path to remove from the beginning of the filenames in the mapping file" type:"path"`
-	VersionCode    string            `help:"Module version code"`
-	VersionName    string            `help:"Module version name"`
+	ApplicationId string            `help:"Module application identifier"`
+	BuildUuid     string            `help:"Module Build UUID"`
+	Path          utils.UploadPaths `arg:"" name:"path" help:"(required) Path to directory or file to upload" type:"path"`
+	ProjectRoot   string            `help:"path to remove from the beginning of the filenames in the mapping file" type:"path"`
+	VersionCode   string            `help:"Module version code"`
+	VersionName   string            `help:"Module version name"`
 }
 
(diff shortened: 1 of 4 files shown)


pkg/upload/android-aab.go
```diff
@@ -104,7 +103,7 @@ func ProcessAndroidAab(apiKey string, androidNdkRoot string, applicationId strin
 
 	if len(fileList) > 0 && err == nil {
 		for _, file := range fileList {
-			err = ProcessAndroidNDK(apiKey, applicationId, androidNdkRoot, "", []string{file}, projectRoot, "", versionCode, versionName, endpoint, failOnUploadError, retries, timeout, overwrite, dryRun)
+			err = ProcessAndroidNDK(apiKey, applicationId, "", "", []string{file}, projectRoot, "", versionCode, versionName, endpoint, failOnUploadError, retries, timeout, overwrite, dryRun)
```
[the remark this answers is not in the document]
joshedney: AndroidNDKRoot check has been moved.

pkg/upload/android-ndk.go
```diff
@@ -155,6 +136,25 @@ func ProcessAndroidNDK(apiKey string, applicationId string, androidNdkRoot strin
...
+				// Find objcopy within NDK path
+				log.Info("Locating objcopy within Android NDK path")
+
+				objCopyPath, err := android.BuildObjcopyPath(androidNdkRoot)
+
+				if err != nil {
+					return err
+				}
+
+				log.Info("Objcopy Path: " + objCopyPath)
```
tomlongridge, line 157: This logging feels a bit over-the-top, can you put this block inside an if-statement that only executes if `objCopyPath` is empty (and declare that variable above)?

Then condense down to a single Info message - "Located objcopy within Android NDK path: {path}".

pkg/upload/android-ndk.go
```diff
@@ -155,6 +137,22 @@ func ProcessAndroidNDK(apiKey string, applicationId string, androidNdkRoot strin
...
+						return err
+					}
+
+					objCopyPath, err = android.BuildObjcopyPath(androidNdkRoot)
+
+					if err != nil {
+						return err
+					}
+
+					log.Info("Located objcopy within Android NDK path:" + androidNdkRoot)
```
tomlongridge, line 154:
```suggestion
					log.Info("Located objcopy within Android NDK path: " + androidNdkRoot)
```

tomlongridge (approved):
LGTM - one tweak

PR #28, opened 2023-04-25
Check if AndroidManifest.xml exists for Android Proguard, NDK and AAB

joshedney:
## Goal

Update the checking around the `AndroidManifest.xml` and only set it if the file exists.

## Design

Follows the pattern set out in the React Native Android section

## Testing

Covered by CI and also manual testing.

### DIFF
--- a/pkg/upload/android-aab.go
+++ b/pkg/upload/android-aab.go
@@ -58,7 +58,6 @@ func ProcessAndroidAab(apiKey string, applicationId string, buildUuid string, pa
 		aabManifestPathExpected := filepath.Join(tempDir, "base", "manifest", "AndroidManifest.xml")
 		if utils.FileExists(aabManifestPath) {
 			aabManifestPath = aabManifestPathExpected
-			log.Info("Found app manifest at: " + aabManifestPath)
 		} else {
 			log.Warn("AndroidManifest.xml not found in AAB file")
 		}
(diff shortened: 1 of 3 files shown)


pkg/upload/android-aab.go
```diff
@@ -48,14 +49,19 @@ func ProcessAndroidAab(apiKey string, applicationId string, buildUuid string, pa
...
 		} else {
 			return fmt.Errorf(path + " is not an AAB file")
 		}
 	}
 
+	if aabManifestPath == "" {
+		aabManifestPathExpected := filepath.Join(tempDir, "base", "manifest", "AndroidManifest.xml")
+		if utils.FileExists(aabManifestPath) {
+			aabManifestPath = aabManifestPathExpected
+			log.Info("Found app manifest at: " + aabManifestPath)
```
tomlongridge, line 61: AABs are a bit different as we'd expect to have a manifest.

I would suggest that we apply the same pattern as the other commands (i.e. only attempt to read it if the file exists and validate that required fields are present before upload). However I would replace this Info message with a Warn if it's not found (i.e. "app manifest not found in AAB") as the path you print here is going to be in the tmp dir and so not very useful.

pkg/upload/android-aab.go
```diff
@@ -48,26 +49,29 @@ func ProcessAndroidAab(apiKey string, applicationId string, buildUuid string, pa
...
```
tomlongridge, line 61: I'd remove this log message as it's in the temp dir, so not a path that means anything to someone with an AAB

PR #32, opened 2023-04-26
feat(create-build): add path option for project dir

tomlongridge:
## Goal

The `create-build` command previously assumes that the git repo it reads for source control parameters is the current directory (from which the command was executed). However I think it makes sense to allow a path to be passed in to the project and this used as the path for git.

## Testing

Manually tested as existing tests don't cover this yet.

### DIFF
--- a/main.go
+++ b/main.go
@@ -1,12 +1,13 @@
 package main
 
 import (
+	"os"
+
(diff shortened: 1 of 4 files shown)


joshedney (approved):
LGTM!

PR #33, opened 2023-04-26
fix(upload-ndk): remove requirment for version info in NDK

tomlongridge:
## Goal

#29 was incorrect for NDK - the version, version code and app ID are only cosmetic in the new NDK endpoint as we use a build ID in the .so file itself to uniquely identify the build. Therefore we don't need to error out if one of them is not present.


joshedney (approved):
LGTM!

PR #34, opened 2023-05-04
[PLAT-10081] Set project root correctly from Android sub folder

joshedney:
## Goal

Set the project root correctly if we're running the command from within the Android sub folder.

## Testing

Tested locally

### DIFF
--- a/pkg/upload/react-native-android.go
+++ b/pkg/upload/react-native-android.go
@@ -76,9 +76,13 @@ func ProcessReactNativeAndroid(apiKey string, appManifestPath string, bundlePath
 		}
 
 		if projectRoot == "" {
-			projectRoot = path
+			if filepath.Base(path) == "android" {
+				projectRoot = filepath.Join(path, "..")
+			} else {
+				projectRoot = path
+			}
 		}
-
+		
 		if appManifestPath == "" {
 			appManifestPathExpected := filepath.Join(buildDirPath, "intermediates", "merged_manifests", variant, "AndroidManifest.xml")
 			if utils.FileExists(appManifestPathExpected) {


pkg/upload/react-native-android.go
```diff
@@ -76,7 +76,11 @@ func ProcessReactNativeAndroid(apiKey string, appManifestPath string, bundlePath
 		}
 
 		if projectRoot == "" {
-			projectRoot = path
+			if filepath.Base(path) == "android" {
+				projectRoot = filepath.Join(path, "..")
+			} else {
+				projectRoot = path
+			}
```
tomlongridge, line 83: I think it would be easier to understand (and less error-prone if the directory just happened to be called "android" for another reason) if we applied this at the same point as we're working out the `buildDirPath` (L37):

```go
		buildDirPath := filepath.Join(path, "android", "app", "build")
		rootDirPath := path
		if !utils.FileExists(buildDirPath) {
			buildDirPath = filepath.Join(path, "app", "build")
			if utils.FileExists(buildDirPath) {
				rootDirPath = filepath.Join(path, "..")
			} else if bundlePath == "" || sourceMapPath == "" {
				return fmt.Errorf("unable to find bundle files or source maps in within " + path)
			}
		}
```

Then you just use this value for project root if not already set:

```suggestion
			projectRoot = rootDirPath
```

tomlongridge: I think you need to amend this check now - it can just assign `projectRoot = rootDirPath`?

joshedney: You are correct. That is my fault I forgot I had the other check further down

pkg/upload/react-native-android.go
```diff
@@ -35,15 +35,19 @@ func ProcessReactNativeAndroid(apiKey string, appManifestPath string, bundlePath
...
 			buildDirPath = filepath.Join(path, "app", "build")
+			rootDirPath = filepath.Join(path, "..")
 			if !utils.FileExists(buildDirPath) {
 				if bundlePath == "" || sourceMapPath == "" {
 					return fmt.Errorf("unable to find bundle files or source maps in within " + path)
 				}
 			}
 		}
 
+		projectRoot = rootDirPath
```
tomlongridge, line 49: We still only want to to do this if it's not specified in the option:

```suggestion
		if projectRoot == "" {
			projectRoot = rootDirPath
		}
```

pkg/upload/react-native-android.go
```diff
@@ -35,15 +35,19 @@ func ProcessReactNativeAndroid(apiKey string, appManifestPath string, bundlePath
 	for _, path := range paths {
 
 		buildDirPath := filepath.Join(path, "android", "app", "build")
+		rootDirPath := path
 		if !utils.FileExists(buildDirPath) {
 			buildDirPath = filepath.Join(path, "app", "build")
+			rootDirPath = filepath.Join(path, "..")
 			if !utils.FileExists(buildDirPath) {
```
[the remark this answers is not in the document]
joshedney: Apologies I was getting confused. I have updated that function to check if the buildDirPaths exits before we set `rootDirPath`

PR #37, opened 2023-05-25
[PLAT-10161] Fix how we check for the AndroidManifest.xml file for Android AAB

joshedney:
## Goal

Update the Android AAB package to ensure we are checking the right path for the AndroidManifest.xml file.

## Changeset

Update the variable we use to check if the AndroidManifest.xml file exists.

## Testing

Tested manually.


lemnik (approved):
LGTM

PR #39, opened 2023-06-07
[PLAT-10142] Add NPM package to install Bugsnag CLI

joshedney:
## Goal

The goal for this PR is to add the `install.js` script to handle the downloading and installing for the correct version of the Bugsnag CLI which will be published as a package to the NPM repository. I've currently got the package in the root of the repo as its only a small file and shouldn't really need anything else adding to it but we may want to move it to its own subfolder for ease of navigation.

Add scripts to the projects `package.json` 

- `bugsnagUpload` - `bugsnag-cli upload react-native-cli` - `npm run bugsnagUpload -- --api-key=YOUR_API_KEY`
- `bugsnagCreateBuild` - `bugsnag-cli create-build` - `npm run bugsnagCreateBuild -- --api-key=YOUR_API_KEY`

## Changeset

Add the foundation for the NPM package so that we can publish it via NPMJS and install the Bugsnag CLI via NPM

## Testing

Tested locally using `verdaccio` 

```
npm install bugsnag-cli --save --registry http://localhost:4873

> bugsnag-cli@1.1.1 postinstall node_modules/bugsnag-cli
> node install.js

Binary downloaded successfully!
npm WARN my-npm-package@1.0.0 No description
npm WARN my-npm-package@1.0.0 No repository field.

+ bugsnag-cli@1.1.1
updated 1 package and audited 15 packages in 5.315s
found 0 vulnerabilities
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,8 +22,5 @@
   ],
   "scripts": {
     "postinstall": "node install.js"
-  },
-  "bin": {
-    "bugsnag-cli": "./node_modules/.bin/bugsnag-cli"
   }
 }
--- a/package.json
+++ b/package.json
@@ -1,11 +1,11 @@
 {
   "name": "bugsnag-cli",
-  "version": "v1.1.1",
+  "version": "1.1.1",
   "description": "BugSnag CLI",
   "main": "install.js",
   "repository": {
     "type": "git",
-    "url": "git+https://github.com/bugsnag/bugsnag-cli.git"
+    "url": "https://github.com/bugsnag/bugsnag-cli"
   },
   "author": "BugSnag",
   "license": "ISC",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "bugsnag-cli",
+  "version": "v1.1.1",
```
gingerbenw, line 3: I think typically this would use `1.1.1` rather than `v1.1.1` - and should we start with a 0 version until we've launched the tool?

```suggestion
  "version": "0.0.1",
```

gingerbenw (approved):

lemnik (approved):
LGTM

PR #44, opened 2023-06-30
[PLAT-10452] Remove hardcoded paths for react native bundle path

joshedney:
## Goal

Allow non-standard variants when not providing the bundle path as a flag to the CLI.

Correctly throw an error when trying to retrieve `apiKey`, `versionName` or `versionCode` from a non-existent android manifest file

## Design

Remove the hardcoded bundle path from the react native upload function and auto detect the variant from the folder structure.

## Testing

Tested locally with multiple variants

### DIFF
--- a/pkg/upload/react-native-android.go
+++ b/pkg/upload/react-native-android.go
@@ -3,6 +3,7 @@ package upload
 import (
 	"fmt"
 	"path/filepath"
+	"strings"
 
 	"github.com/bugsnag/bugsnag-cli/pkg/android"
 	"github.com/bugsnag/bugsnag-cli/pkg/log"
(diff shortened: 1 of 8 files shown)


pkg/upload/react-native-android.go
```diff
@@ -19,11 +20,12 @@ type ReactNativeAndroid struct {
 	ProjectRoot  string            `help:"path to remove from the beginning of the filenames in the mapping file" type:"path"`
 	SourceMap    string            `help:"Path to the source map file" type:"path"`
 	Variant      string            `help:"Build type, like 'debug' or 'release'"`
-	Version      string            `help:"The version of the application."`
+	Version      string            `help:"(depricated) The version name of the application."`
```
tomlongridge, line 23:
```suggestion
	Version      string            `help:"(deprecated) The version name of the application."`
```

pkg/upload/react-native-android.go
```diff
@@ -105,41 +120,46 @@ func ProcessReactNativeAndroid(apiKey string, appManifestPath string, bundlePath
...
+			} else {
+				appManifestPath = ""
 			}
 		}
 
-		if appManifestPath != "" && (apiKey == "" || version == "" || versionCode == "") {
-
-			manifestData, err := android.ParseAndroidManifestXML(appManifestPath)
+		if apiKey == "" || versionName == "" || versionCode == "" {
+			if appManifestPath != "" {
```
tomlongridge, line 129: Was this causing an issue? I believe this was the intended approach - we read the manifest if any of the data is missing and only if the manifest is found. Otherwise we allow `BuildReactNativeAndroidUploadOptions` to generate the error.

tomlongridge (approved):

PR #51, opened 2023-07-20
[PLAT-10636]  Add version flag

joshedney:
## Goal

Adjust the build scripts to embed the version number into the binary and to add a CLI flag to retrieve this version number. 

This will help users identify which version of the CLI they are using.

## Changeset

Add  a new script, `./script/bump-version.sh`, to manage setting of the version number across `main.go`, package.json` and the `CHANGELOG.md`.

## Testing

Covered via CI


makefile
```diff
@@ -13,6 +13,8 @@ endif
 endif
 endif
 
+PACKAGE_VERSION :=$(shell cat package.json | jq -r .version)
```
tomlongridge, line 16: This works, but it's a different approach to our notifiers and so I think it might trip us up if we're not careful. (Also, the source of truth being inside the package.json is a bit odd for primarily a Go package.)

Can you take a look at the `Makefile` and `bump-version.sh` script in bugsnag/bugsnag-android-performance? I think that would be our best pattern to follow - the version is either derived from the branch name or entered manually and it sets the values in all the places in the repo that it needs to, including the changelog.

In this case you're using the version in the `Makefile` itself, so we would want to refactor so that it either sets the value in `main.go` source or create a file that just stores the version (e.g. https://github.com/bugsnag/bugsnag-ruby/blob/master/VERSION)

scripts/bump-version.sh
```diff
@@ -0,0 +1,17 @@
...
+else
+  echo "$BRANCH does not appear to be a release branch, please specify VERSION manually"
+  echo "$(basename $0) <version-number>"
+  exit 1
+fi
+
+echo Bumping the version number to $VERSION
+sed -i '' "s/package_version = \".*\"/package_version = \"$VERSION\"/" main.go
+sed -i '' "s/\"version\": \".*\"/\"version\": \"$VERSION\"/" package.json
+sed -i '' "s/## TBD/## $VERSION ($(date '+%Y-%m-%d'))/" CHANGELOG.md
```
tomlongridge, line 17: The changelog style is currently h1 (`#`) for the version/date line - in fact, most changelogs have a single h1 `# Changelog` and then h2's for each version. Might be worth matching that style rather than changing this.

makefile
```diff
@@ -13,6 +13,7 @@ endif
 endif
 endif
 
+
```
tomlongridge, line 16: Can you add a `bump` target:

```
bump:
ifneq ($(shell git diff --staged),)
	@git diff --staged
	@$(error You have uncommitted changes. Push or discard them to continue)
endif
	@./scripts/bump-version.sh $(VERSION)
```

tomlongridge: This allows you to do `make VERSION=1.2.3 bump` like other repos.

Also, can you update the CONTRIBUTING.md guide?

scripts/bump-version.sh
tomlongridge: > Might be worth matching that style rather than changing this.

I'd rather the changelog matched our normal style - each version being an h2, with a single h1 "Changelog"

CHANGELOG.md
```diff
@@ -1,48 +1,49 @@
-# 2.0.0 (TBD)
+# Change Log
```
tomlongridge, line 1:
```suggestion
# Changelog

```

tomlongridge (approved):

PR #52, opened 2023-07-20
[PLAT-10642] Remove deprecated CLI options

joshedney:
## Goal

Remove deprecated `appVersion`, `appVersionCode` and `appBundleVersion` options from the CLI to make way for the `--version` flag 

## Changeset

Remove any CLI option from the CLI that has been marked as deprecated.

## Testing

Covered via CI


CHANGELOG.md
```diff
@@ -1,32 +1,48 @@
-# Changelog
+# 1.3.0 (TBD)
```
tomlongridge, line 1: If we're removing the flags, then this will need to be major version bump.

tomlongridge (approved):

PR #53, opened 2023-07-28
[PLAT-10637] Rework integration tests

joshedney:
## Goal

Rework the tests to add better coverage for the currently supported source map uploads.

For each currently supported technology (Dart, Android AAB, NDK, Proguard and React Native Android) we mock the expected fils paths so that we can test the automatic detection for required files. These files are also used to test the escape hatches for the commands by allowing us to provide values for each flag option. 

React Native Android has multiple versions (0.69, 0.70 and 0.72) due to the change in file/folder structure between the different versions. 

Also included in this PR is a set of `base-fixtures` which is used to generate new test fixtures. These test fixtures are then used to run the tests against to ensure that the automatic file/folder detection works against an actual project. 

When a new version of the test fixtures are required, these `base-fixtures` can be updated and used to run tests against. We can then update the mock file/folders as required should a change occur.

## Changeset

Added integration test for the currently supported source maps (react-native android, android ndk/proguard/aab and dart) to ensure that automatic file detection and manual detection works.  Uploads are verified by sending the requests to mock endpoints within Maze Runner.

## Testing

Covered by CI


tomlongridge (changes requested):
Can you remove all the images (e.g. the flutter icons) to avoid bloating the repo?

joshedney:
> Can you remove all the images (e.g. the flutter icons) to avoid bloating the repo?

All test fixture images have been removed.

tomlongridge (approved):

PR #54, opened 2023-07-31
Build ID derived from .dex signatures

lemnik:
## Goal
Use a build ID derived from the `.dex` file signatures when no Build ID is provided (either as a command-line option, or within the `AndroidManifest.xml`).

## Design
When no Build UUID is provided on the command line, or found in the `AndroidManifest.xml` of an Android `.aab` file the CLI will now inspect the `classes[N].dex` files for their [SHA-1 signatures](https://source.android.com/docs/core/runtime/dex-format#header-item) and XOR them together to produce a Build UUID that ties the mapping files back to the exact AAB file.

## Testing
Manual testing, a new unit test and a new end-to-end test.


joshedney (approved):
LGTM - Just need to update the changelog

PR #55, opened 2023-08-02
[PLAT-10654] Add --dry-run flag to upload commands

joshedney:
## Goal

Validate but not upload source maps for all supported languages.

## Changeset

Add the `--dry-run` flag to all upload commands to process given source maps but not upload them. 
Removed the dry run log message from the `AAB` command to reduce the number of log messages

## Testing

Tested manually, CI tests to be added during testing rework.

### DIFF
--- a/pkg/server/request.go
+++ b/pkg/server/request.go
@@ -94,9 +94,9 @@ func ProcessRequest(endpoint string, uploadOptions map[string]string, fileFieldD
 			return fmt.Errorf("%s : %s", res.Status, string(b))
 		}
 
-		log.Success(fileName + " Processed")
+		log.Success("Uploaded " + fileName)
 	} else {
-		log.Success("(dryrun) " + fileName + " Processed")
+		log.Success("(dryrun) Uploaded " + filepath.Base(fileName))
 	}
 
 	return nil
--- a/pkg/server/request.go
+++ b/pkg/server/request.go
@@ -66,7 +66,7 @@ func SendRequest(request *http.Request, timeout int) (*http.Response, error) {
 }
 
 // ProcessRequest - Builds and sends file requests to the API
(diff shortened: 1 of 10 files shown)


pkg/upload/android-proguard.go
```diff
@@ -27,6 +27,10 @@ func ProcessAndroidProguard(apiKey string, applicationId string, appManifestPath
 	var appManifestPathExpected string
 	var err error
 
+	if dryRun {
+		log.Info("Performing dry run - no files will be uploaded")
+	}
```
tomlongridge, line 32: Does dryRun now apply to all commands? Could we just put a message at the very start, top-level? I guess it just needs to be worded more generally – "no data will be sent to BugSnag"?

pkg/server/request.go
```diff
@@ -69,32 +66,37 @@ func SendRequest(request *http.Request, timeout int) (*http.Response, error) {
...
+
+		if !statusOK {
+			return fmt.Errorf("%s : %s", res.Status, string(b))
+		}
 
-	if !statusOK {
-		return fmt.Errorf("%s : %s", res.Status, string(b))
+		log.Success("Uploaded " + filepath.Base(fileName))
+	} else {
+		log.Success("(dryrun) Uploaded " + filepath.Base(fileName))
```
lemnik, line 99: I would still be nervous about this wording - "Uploaded" still sounds like something happened.

I would personally prefer something that made it absolutely clear nothing had happened. Something like:

```go
log.Success("(dryrun) Skipping upload of " + filepath.Base(fileName) + " to " + endpoint)
```

joshedney: Thoughts @tomlongridge ?

tomlongridge: Agreed - maybe also `.Info` log to underscore it?

Are you going to do equivalent for `Create.go`?

tomlongridge (approved):

PR #59, opened 2023-10-12
[PLAT-10966] Install script version handling

joshedney:
## Goal

Bypass the rate limiting of the Github API when installing via the `install.sh` script and allow users to specify the version they wish to install via a command line flag.

Example:

`./install.sh --version=1.2.2`

Output:

```
==> This script will install:
/Users/josh.edney/.local/bugsnag/bin/bugsnag-cli
==> Downloading and installing Bugsnag CLI...
############################################## 100.0%
==> Installation successful!
```

## Changeset

Instead of using the github API to search for the latest release of the CLI we set the latest version using the `bump.sh` script within the `install.sh` script. We have also added an option to allow users to specify the version they wish to install using a `--version=x.x.x` flag

## Testing

Tested locally but we could add this to the CI run if we wanted. Currently the CLI used is built from source.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -83,7 +83,7 @@ file_not_grpowned() {
 display_help() {
   echo ""
   cat <<EOS
-  Usage: ./$0
+  Usage: ./install.sh
 
   Flags:
     --help                          Display help and usage information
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -80,7 +80,36 @@ file_not_grpowned() {
   [[ " $(id -G "${USER}") " != *" $(get_group "$1") "* ]]
 }
 
-BUGSNAG_CLI_GIT_REMOTE="https://api.github.com/repos/bugsnag/bugsnag-cli/releases/latest"
+display_help() {
+  echo ""
+  cat <<EOS
+  Usage: ./$0
```
tomlongridge, line 86: This outputs the following when you `./` the script:

```
  Usage: ././install.sh
```

Any reason not remove the "./" in the script or just to hard-code `./install.sh`?

install.sh
```diff
@@ -136,18 +165,7 @@ ohai "Downloading and installing Bugsnag CLI..."
...
-    cut -d : -f 2,3 |
-    tr -d \" |
-    xargs)
-
-  if [[ ! -n ${DOWNLOAD_URL} ]]; then
-    abort "Failed to get download URL from ${BUGSNAG_CLI_GIT_REMOTE}"
-  fi
-
-  execute "curl" "-#" "-L" "${DOWNLOAD_URL}" "-o" "${BUGSNAG_CLI_PREFIX}/bin/bugsnag-cli"
+  execute "curl" "-#" "-L" "https://github.com/bugsnag/bugsnag-cli/releases/download/v${VERSION}/${UNAME_MACHINE}-${OS_NAME}-bugsnag-cli" "-o" "${BUGSNAG_CLI_PREFIX}/bin/bugsnag-cli"
```
tomlongridge, line 168: Is there a way to check whether the download succeeded? If you enter a non-existent version, it's not clear from the output that nothing happened.

tomlongridge (approved):

PR #61, opened 2023-10-17
Proguard dex signatures

lemnik:
## Goal
Allow progaurd `mapping.txt` files to be uploaded with stable (`.dex` derived) build ids.

## Changes
`upload android-proguard` now has a `--dex-files` option which can specify:

- a list of `classes.dex` files
- a directory containing the `.dex` files
- any mixture of the above

A new `create-android-build-id` top-level command can be used to generate and output the stable build id without performing an upload.

## Design
Split the searching for `classes.dex` files from the signature generation so that the `--dex-files` option could share the same signature generation logic.

## Testing
Manual testing.


joshedney (approved):
LGTM! Do we need to update the changelog ?

PR #65, opened 2023-11-27
[PLAT-11295] Gather build information from environment and provided files

joshedney:
## Goal

Allow create build to gather settings for several locations to be merged and validated before calling the process function. 

## Testing

Covered by new CI tests


pkg/options/options.go
```diff
@@ -0,0 +1,60 @@
...
+	Provider          string            `help:"The name of the source control provider that contains the source code for the build."`
+	Repository        string            `help:"The URL of the repository containing the source code being deployed."`
+	Revision          string            `help:"The source control SHA-1 hash for the code that has been built (short or long hash)"`
+	Path              utils.Paths       `arg:"" name:"path" help:"Path to the project directory" type:"path" default:"."`
+	VersionName       string            `help:"The version of the application." xor:"app-version,version-name"`
+	VersionCode       string            `help:"The version code for the application (Android only)." xor:"app-version-code,version-code"`
+	BundleVersion     string            `help:"The bundle version for the application (iOS only)." xor:"app-bundle-version,bundle-version"`
+	AutoAssignRelease bool              `help:"TBC"`
+	AppManifest       utils.Path        `help:"TBC"`
+	AndroidAab        utils.Path        `help:"TBC"`
```
tomlongridge, line 59: Can the Android-specific options be nested into a `build.Android` struct (or similar) to logically separate them? Once we have iOS options and potential future ones, I think keeping them separate will be useful.

pkg/build/create.go
```diff
@@ -36,41 +23,16 @@ type Payload struct {
...
-			Provider:   repoInfo["provider"],
-			Repository: repoInfo["repository"],
-			Revision:   repoInfo["revision"],
-		},
-		Metadata:         metadata,
-		AppVersion:       version,
-		AppVersionCode:   versionCode,
-		AppBundleVersion: bundleVersion,
+		return err
 	}
```
tomlongridge, line 34: This feels like it would be better in the caller

main.go
```diff
@@ -244,11 +211,32 @@ func main() {
...
+		if androidManifestPath == "" {
+			androidManifestPath = string(commands.CreateBuild.AppManifest)
 		}
 
+		if androidManifestPath != "" {
+			ManifestBuildOptions := build.PopulateFromAndroidManifest(androidManifestPath)
+			BaseOptions = build.TheGreatMerge(BaseOptions, ManifestBuildOptions)
+		}
+
+		FinalMerge := build.TheGreatMerge(UserBuildOptions, BaseOptions)
```
tomlongridge, line 238: Can all this go into a func in `build`? So that all that's left on the top level is:

1. Create build info
2. Validate build info
3. Get the endpoint URL
4. Call the process function

pkg/options/options.go
joshedney: Should be possible.

pkg/android/get-android-info.go
```diff
@@ -0,0 +1,238 @@
...
+	rawAabManifestData := &proto_messages.XmlNode{}
+
+	err = proto.Unmarshal(content, rawAabManifestData)
+
+	if err != nil {
+		return nil, err
+	}
+
+	for _, data := range rawAabManifestData.GetElement().GetAttribute() {
+		if data.ResourceId == 16843291 || data.Name == "versionCode" {
```
lemnik, line 135: It might be worth having a comment to document where these numbers come from:
https://developer.android.com/reference/android/R.attr#versionCode

Or better yet: move them to a constant and put the comment there?

pkg/android/get-android-info.go
```diff
@@ -0,0 +1,238 @@
...
+			aabManifestData["versionCode"] = data.GetValue()
+			continue
+		}
+
+		if data.Name == "package" {
+			aabManifestData["applicationId"] = data.GetValue()
+			continue
+		}
+
+		if data.ResourceId == 16843292 || data.Name == "versionName" {
```
lemnik, line 145: https://developer.android.com/reference/android/R.attr#versionName

main.go
```diff
@@ -244,11 +211,32 @@ func main() {
...
+		}
+
+		if androidManifestPath == "" {
+			androidManifestPath = string(commands.CreateBuild.AppManifest)
 		}
 
+		if androidManifestPath != "" {
+			ManifestBuildOptions := build.PopulateFromAndroidManifest(androidManifestPath)
+			BaseOptions = build.TheGreatMerge(BaseOptions, ManifestBuildOptions)
+		}
```
lemnik, line 236: This feels like it should be in it's own function?

pkg/build/gather-build-info.go
```diff
@@ -0,0 +1,152 @@
...
+
+	if opts.AppVersion == "" {
+		return fmt.Errorf("Missing app version, please provide this via the command line options")
+	}
+
+	if opts.SourceControl.Repository == "" || opts.SourceControl.Revision == "" {
+		return fmt.Errorf("Missing Source Control Repository or Revision")
+	}
+
+	return nil
```
tomlongridge, line 59: We should also validate that provider (if given) is one of the permitted values:

* github
* github-enterprise
* bitbucket
* bitbucket-server
* gitlab
* gitlab-onpremise

pkg/build/gather-build-info.go
```diff
@@ -0,0 +1,152 @@
...
+		return fmt.Errorf("no API key provided")
+	}
+
+	if opts.AppVersion == "" {
+		return fmt.Errorf("Missing app version, please provide this via the command line options")
+	}
+
+	if opts.SourceControl.Repository == "" || opts.SourceControl.Revision == "" {
+		return fmt.Errorf("Missing Source Control Repository or Revision")
+	}
```
tomlongridge, line 57: Are these fields definitely required for the Build API or can the whole of `sourceControl` be omitted?

If this is true, we should split the validation and resulting error messages.

pkg/build/gather-build-info.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+func (opts CreateBuildInfo) Validate() error {
+	if opts.ApiKey == "" {
+		return fmt.Errorf("no API key provided")
+	}
+
+	if opts.AppVersion == "" {
+		return fmt.Errorf("Missing app version, please provide this via the command line options")
```
tomlongridge, line 52: Think we need a bit more consistency in these validation messages - casing, phrasing and also whether we provide a useful nudge. I'm not sure on this last point whether you could do that in a platform-agnostic way, so I'd be tempted to remove entirely.

pkg/build/process-create-build.go
```diff
@@ -0,0 +1,66 @@
...
+	if !dryRun {
+		res, err := server.SendRequest(req, 300)
+
+		if err != nil {
+			return fmt.Errorf("error sending file request: %w", err)
+		}
+
+		b, err := io.ReadAll(res.Body)
+
+		if strings.Contains(string(b), "Source control provider is missing") {
```
tomlongridge, line 50: Can we just read the `warnings` field of the response and avoid this specific logic?

pkg/build/process-create-build.go
```diff
@@ -0,0 +1,66 @@
...
+func ProcessBuildRequest(buildOptions CreateBuildInfo, endpoint string, dryRun bool) error {
+
+	buildPayload, err := json.Marshal(buildOptions)
+
+	if err != nil {
+		log.Error("Failed to create build information payload: "+err.Error(), 1)
+	}
+
+	prettyBuildPayload, _ := utils.PrettyPrintJson(string(buildPayload))
+	log.Info("Build information: \n" + prettyBuildPayload)
```
tomlongridge, line 35: Is this in-keeping with logging elsewhere? Feels like it might be too low level - or use debug log?

pkg/build/process-create-build.go
```diff
@@ -0,0 +1,66 @@
...
+
+	req, _ := http.NewRequest("POST", endpoint, bytes.NewBuffer(buildPayload))
+
+	req.Header.Add("Content-Type", "application/json")
+
+	if !dryRun {
+		res, err := server.SendRequest(req, 300)
```
tomlongridge, line 42: Not using the configured timeout

pkg/build/process-create-build.go
```diff
@@ -0,0 +1,66 @@
...
+
+		if err != nil {
+			return fmt.Errorf("error sending file request: %w", err)
```
tomlongridge, line 45: Could more of `server.ProcessRequest` be abstracted in `SendRequest` – the request and response error check and status code; also dry run? Then it needn't be repeated here.

pkg/utils/get-repository-info.go
```diff
@@ -51,3 +51,25 @@ func GetCommitHash() string {
...
+	}
+
+	if repoHash == "" {
+		repoHash = GetCommitHash()
+	}
+
+	repoInfo["revision"] = repoHash
+
+	return repoInfo
+}
```
tomlongridge, line 75: I don't think this is used anymore?

pkg/build/gather-build-info.go
```diff
@@ -0,0 +1,152 @@
...
+	return CreateBuildInfo{
+		ApiKey:         apiKey,
+		AppVersionCode: androidData.VersionCode,
+		AppVersion:     androidData.VersionName,
+	}
+}
+
+func TheGreatMerge(p1 CreateBuildInfo, p2 CreateBuildInfo) CreateBuildInfo {
+	return p1.Override(p2)
+}
```
tomlongridge, line 121: I think it's actually more readable without this abstraction:

```
x.Override(y)
```

pkg/android/get-android-info.go
```diff
@@ -0,0 +1,245 @@
...
+		return false, err
+	}
+
+	// Check for the protobuf identifier
+	protobufIdentifier := []byte{0x0A} // Example: Protobuf identifier often starts with 0x0A
+	return bytes.HasPrefix(buffer, protobufIdentifier), nil
+}
+
+// Function to extract the android manifest file from a given aab file
+func GetAndroidManifestFileFromAAB(path string) (string, error) {
```
tomlongridge, line 67: Can this be shared with the existing AAB functionality? (Could be a later piece of work, but long-term it doesn't feel like we should have this duplicated.)

pkg/android/get-android-info.go
```diff
@@ -0,0 +1,245 @@
...
+
+				return outputPath, nil
+			}
+		}
+	}
+	return "", fmt.Errorf("no aab file")
+}
+
+// GetAndroidXMLData - Pulls information from a human-readable xml file into a struct
+func GetAndroidXMLData(manifestFile string) (*AndroidManifestData, error) {
```
tomlongridge, line 109: Is there any golang naming convention or syntax for marking a function as internal to the package? It would make reading the package a little easier.

pkg/android/get-android-info.go
```diff
@@ -0,0 +1,245 @@
...
+				Value: []string{aabManifestData["apiKey"], aabManifestData["buildUuid"]},
+			},
+		},
+	}
+
+	return manifestData, nil
+}
+
+func BuildAndroidInfo(path string) (*AndroidManifestData, error) {
+	//	Workout if its xml or protobuff
```
tomlongridge, line 201: Not sure these comments add anything - the functions they call are self-documenting. Perhaps the func needs an overall comment to summarise intent?

pkg/build/gather-build-info.go
joshedney: According to [this](https://bugsnagbuildapi.docs.apiary.io/#reference/0/build/notify-of-a-build). `repository` and `revision` are. Provider looks to be inferred on the API side which is why there is no validation within the CLI on that option.

pkg/build/process-create-build.go
joshedney: Even though this looks like a new addition. It in fact has not changed. Only the file name has changed hence why it look s like a new addition. It was a feature that was previously requested but happy to adjust if we feel like it is no longer needed.

pkg/build/gather-build-info.go
tomlongridge: Would you mind trying a few combinations against the API? I'm a little doubtful if that is strictly the case as I'd have thought you'd be able to create a release without source control information. Perhaps they're required if you want to provide any `sourceControl` fields?

pkg/build/process-create-build.go
tomlongridge: Fair point - it was probably me that requested it before :) Leave it as-is for now then.

pkg/build/gather-build-info.go
joshedney: My mistake. You are correct. They're only required if you pass the `sourceControl` element

pkg/android/get-android-info.go
joshedney: Yes long term it will be shared.

pkg/android/get-android-info.go
joshedney: Usually internal functions are `internalFunction` and public functions are `PublicFunction` or we could move them to an `internal` package so that it becomes `internal.MyFunction`

pkg/build/gather-build-info.go
```diff
@@ -0,0 +1,147 @@
...
+
+func PopulateFromPath(path string) CreateBuildInfo {
+	return CreateBuildInfo{
+		ApiKey:           "",
+		AppVersionCode:   "",
+		AppBundleVersion: "",
+		SourceControl: SourceControl{
+			Provider:   "",
+			Repository: utils.GetRepoUrl(path),
+			Revision:   utils.GetCommitHash(),
```
tomlongridge, line 84: There's somewhat of an assumption that they are using git. Does this fail OK if that is not the case? ... or if git isn't on the path?

I think it would also be good to rename the functions to make it clear this is git only.

pkg/build/gather-build-info.go
tomlongridge: Discussed IRL - we'll let the API do the validation.

tomlongridge (approved):

PR #68, opened 2023-12-13
[PLAT-11330] Adjust how we handle multiple file uploads and failures for Android NDK/AAB

joshedney:
## Goal

Ensure that we properly handle failures with multiple file uploads for Android NDK and AAB when passing the `--fail-on-upload-error` CLI flag.

## Changeset

Rather than process each NDK file separately when dealing with an AAB file, we will pass an array of strings so that it can process them as one group enabling us to utilise the `--fail-on-upload-error` flag correctly. 

Adjust how we check if we're dealing with multiple files when checking for `failOnUploadError` to ensure that we still error if we're dealing with a single file.

## Testing

Covered by CI


lemnik (approved):
LGTM

PR #69, opened 2023-12-15
Correct the Android NDK / Proguard help

lemnik:
## Goal
Let users know that `upload android-ndk` will upload the NDK symbols, and not the Proguard mapping files .

## Changes
The help text for `upload android-ndk` now indicates that the command will upload the NDK symbols, and not the Proguard mapping files (as previously advertised). Similarly `upload android-proguard` no longer suggests it will upload the NDK symbols.

As a point of clarity the `upload android-proguard` help text now shows that Proguard & R8 are treated as equivalent by the CLI (as they both produce a `mapping.txt` in the same locations).

## Testing
Tested manually by running the CLI with and commands


joshedney (approved):
Good catch! I've updated the base to point to the release branch

PR #70, opened 2024-01-23
[HOTFIX] Ensure `--retries` is handled correctly during uploads

joshedney:
## Goal

Ensure that when using the `--retries` argument on the CLI that if the upload fails the upload is retried the number of times equal to the value passed to the CLI argument. 

## Changeset

Add a new function to house and process the retry logic for sending requests to the upload endpoint.

## Testing

Covered by CI


lemnik (approved):
LGTM

PR #72, opened 2024-01-31
[PLAT-11493] xcodebuild utility function updates

jasonMcCullough:
## Goal

The getXcodeSchemes and getXcodeBuildSettings functions currently take a workspacePath argument which can be replaced by a path instead to make them a bit more flexible so they can be used to check if a workspace directory is passed, apply -workspace flag, if a project directory is passed then apply -project flag. Either path is empty, then just run xcodebuild without either flag as you would when running in the root of an xcode project.

## Design

To partially set things up for the new <path> behaviour. As explained in my design [here](https://smartbear.atlassian.net/wiki/spaces/EN/pages/3891986527/PLAT-11454+ED+-+iOS+dSYM+uploads+BugSnag+CLI#New-%3Cpath%3E-behaviour-explained).

## Changeset

The getXcodeSchemes and getXcodeBuildSettings functions

## Testing

N/A - I'll test everything with MazeRunner at the end in the main branch.

### DIFF
--- a/pkg/ios/xcodebuild-utils.go
+++ b/pkg/ios/xcodebuild-utils.go
@@ -18,23 +18,29 @@ type XcodeBuildSettings struct {
 	DsymName              string `mapstructure:"DWARF_DSYM_FILE_NAME"`
 }
 
-// IsSchemeInWorkspace checks if a scheme is in a given workspace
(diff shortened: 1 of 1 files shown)


joshedney (approved):
LGTM!

PR #73, opened 2024-01-31
[PLAT-11494] New default scheme behaviour

jasonMcCullough:
## Goal

Determines the default scheme in absence of a provided scheme.

## Design

Details in my design can be found [here](https://smartbear.atlassian.net/wiki/spaces/EN/pages/3891986527/PLAT-11454+ED+-+iOS+dSYM+uploads+BugSnag+CLI#New-code-to-be-implemented-into-the-BugSnag-CLI).

Summary:

If no scheme provided:
1) If one scheme is found, use by default.
2) If no schemes are found, error and direct user to define with `--scheme`
3) If multiple schemes are found, error and direct user to define with `--scheme`

## Changeset

New `GetDefaultScheme` function and implemented where applicable in RN iOS for now. Will also be used for regular dSYM upload functionality too when the time comes.

## Testing

Added `--scheme` to tests to ensure when multiple schemes were detected that the correct one is chosen post-rework.

### DIFF
--- a/features/base-fixtures/rn0_69/Gemfile
+++ b/features/base-fixtures/rn0_69/Gemfile
@@ -3,4 +3,4 @@ source 'https://rubygems.org'
 # You may use http://rbenv.org/ or https://rvm.io/ to install and use this version
 ruby '>= 2.6.10'
 
-gem 'cocoapods', '1.14.3'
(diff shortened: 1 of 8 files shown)


joshedney (approved):
LGTM!

PR #74, opened 2024-02-01
[PLAT-11496] A new function for gatekeeping the value of the path argument

jasonMcCullough:
## Goal

This should help manage the passed value of `<path>` since it will ultimately aid in deciding which value project root should be set to (if not already set explicitly by the user).

## Design

We've approached this in a way that helps us accurately send a project root to the upload server.

## Changeset

A new function `ProcessPathValue` which serves the above purpose.

Also tweaks to pre-existing util functions to take the value of projectRoot. Can be used to determine where to run the command in scenarios where `-workspace`or `-project` flag are not provided.

## Testing

Tested common use cases outlined in my design ([see here](https://smartbear.atlassian.net/wiki/spaces/EN/pages/3891986527/PLAT-11454+ED+-+iOS+dSYM+uploads+BugSnag+CLI#New-%3Cpath%3E-behaviour-explained)). This function is meant to do a lot of the heavy lifting for us in order to accurately determine the value of project root. Includes common edge case where `<path>` may be a dSYM.


jasonMcCullough:
@tomlongridge I think I've accurately depicted our expected common use cases in my test `TestProcessPathValue`, but feel free to suggest any edge cases I might have missed that are worth adding. At this point I'm sort of _imagining_ that these function arguments are derived from `<path>` and `--project-root` but the process of actually retrieving them is not implemented yet. Just trying to get util groundwork sorted out first and then actual upload logic will come at the latter half of the project 😄

jasonMcCullough:
@joshedney I think this is probably enough test coverage for this, what do you think and I can merge this and close my ticket?

joshedney (approved):
LGTM!

PR #75, opened 2024-02-02
[PLAT-11495] Core upload functionality for dSYM iOS

jasonMcCullough:
## Goal

A means of facilitating dSYM uploads in BugSnag CLI. (Less the functionality that is to be implemented in [PLAT-11497](https://smartbear.atlassian.net/browse/PLAT-11497))

## Design

The boilerplate functionality is based on pre-existing uploads and shares similarities to RN iOS for validation etc.

## Changeset

1) Added a new set of arguments specific to dSYM.
2) Some util tweaks that have been made compatible with RN iOS as well. Which includes a derivedFrom return value which is useful when displaying where the scheme is derived from (either from `<path>` or `--project-root` when it can be either in some scenarios)
3) Ability to explicitly define `--dsym-path` which should follow same order of precedence as setting `--project-root`. (See design for details)
4) Leveraging dwarfdump for fetching dwarf/dsym file information for uploading. Acts a means of ensuring only valid artifacts are teed up for upload as well.
5) A basic `.zip` compression check and a means of using the temporary location for uploading.

## Testing

Automation testing - Minor changes to pre-existing unit tests.
Maze Runner - Will likely get around to this near the end and tracked via this [PLAT-11498](https://smartbear.atlassian.net/browse/PLAT-11498)

[PLAT-11497]: https://smartbear.atlassian.net/browse/PLAT-11497?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PLAT-11498]: https://smartbear.atlassian.net/browse/PLAT-11498?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/upload/dsym.go
```diff
@@ -0,0 +1,128 @@
...
+
+		// If scheme is set explicitly, check if it exists
+		if scheme != "" {
+			_, err = ios.IsSchemeInPath(path, scheme, projectRoot)
+			if err != nil {
+				return err
+			}
+		} else {
+			// If the scheme is not set explicitly, try to find it
+			scheme, err = ios.GetDefaultScheme(path, projectRoot)
```
tomlongridge, line 60: Do we only need `scheme` if the dsym path is not given? If so, we shouldn't error if we can't find a default scheme as that won't be needed.

pkg/upload/dsym.go
```diff
@@ -0,0 +1,128 @@
...
+				return err
+			}
+
+			// Build the dsymPath from build settings
+			uploadInfo.DsymPath = buildSettings.ConfigurationBuildDir + "/" + buildSettings.DsymName + "/Contents/Resources/DWARF"
+
+			filesFound, _ := os.ReadDir(uploadInfo.DsymPath)
+			switch len(filesFound) {
+			case 0:
+				return errors.Errorf("No files found in location '%s'", uploadInfo.DsymPath)
```
tomlongridge, line 80: We need to be careful with this message – the user hasn't explicitly told us to look here, so we should say something more like "No dSYM files found in _expected_ location ..."

Also, should we suggest running with a dsym path or provide a scheme name?

pkg/upload/dsym.go
```diff
@@ -0,0 +1,128 @@
...
+			case 1:
+				uploadInfo.DsymPath = filepath.Join(uploadInfo.DsymPath, filesFound[0].Name())
+			default:
+				return errors.Errorf("Multiple files found in location '%s'", uploadInfo.DsymPath)
```
tomlongridge, line 84: Do you think this is a plausible case or just a technical possibility?

pkg/upload/dsym.go
jasonMcCullough: Ah - Yeah we're only dealing with the scheme if the dsym path is not given. Maybe I can just gate this whole block of code behind a condition e.g. `if uploadInfo.DsymPath == ""`

pkg/upload/dsym.go
jasonMcCullough: Yeah perhaps having an option to explicitly set `--dsym-path`? Then I can handle it like we do with the order of precedence that we already do with `--project-root`

pkg/upload/dsym.go
jasonMcCullough: I was torn with this actually, I don't think it's actually possible but I put it there just in case

jasonMcCullough: Just thought I'd have a look at [this](https://github.com/bugsnag/bugsnag-dsym-upload/blob/47fee06af3a6d92e8aafd3c9d3453649ded03e13/bin/bugsnag-dsym-upload#L170C34-L170C46) again. Looks like we were checking all files in the directory and uploading those that produced a UUID via dwarfdump, so I'll just do the same I think to be safe

pkg/upload/dsym.go
```diff
@@ -0,0 +1,176 @@
...
+				}
+				log.Info("Using alternative dSYM path: " + dsymPath)
+			}
+
+		} else {
+			// Use the dsymPath from the command line, check if it's zipped and unzip it if necessary
+			var extractedLocation string
+			extractedLocation, err = utils.ExtractFile(dsymPath, "dsym")
+			if err != nil {
+				log.Warn(utils.DisplayBlankIfEmpty(dsymPath) + " is not a zip file or directory")
```
tomlongridge, line 108: Does `ExtractFile` return `nil` if the target isn't a zip file? If so, when might `err` be set? I was a bit confused by the message (and the fact that it's a Warn) as I wouldn't expect a warning to cause an early return and according to the comments above, we are expecting some cases where it's not a zip file.

pkg/upload/dsym.go
```diff
@@ -0,0 +1,176 @@
...
+		dsyms, err = ios.GetDsymsForUpload(dsymPath)
+		if err != nil {
+			return err
+		}
+
+		for _, dsym := range *dsyms {
+			dsymInfo := "(UUID: " + dsym.UUID + ", Name: " + dsym.Name + ", Arch: " + dsym.Arch + ")"
+			log.Info("Uploading dSYM " + dsymInfo)
+			// If the Info.plist path is defined and we still don't know the apiKey or verionName, try to extract them from it
+			if plistPath != "" && (apiKey == "" || versionName == "") {
```
tomlongridge, line 127: I don't believe this needs to be done for each dsym - just once?

pkg/ios/dwarfdump-utils.go
```diff
@@ -0,0 +1,82 @@
...
+				if strings.HasSuffix(file.Name(), ".zip") {
+					log.Info("Attempting to unzip " + file.Name() + " before proceeding to upload")
+					path, _ = utils.ExtractFile(filepath.Join(path, file.Name()), "dsym")
+
+					if path != "" {
+						log.Info("Unzipped " + file.Name() + " to " + path + " for uploading")
+					}
+				}
+
+				cmd := exec.Command("dwarfdump", "-u", strings.TrimSuffix(file.Name(), ".zip"))
```
tomlongridge, line 47: Should this use the `DWARFDUMP` constant?

pkg/ios/dwarfdump-utils.go
```diff
@@ -0,0 +1,82 @@
...
+
+	switch len(filesFound) {
+	case 0:
+		return nil, errors.Errorf("No dSYM files found in expected location '%s'", path)
+	default:
+		if isDwarfDumpInstalled() {
+			for _, file := range filesFound {
+				if strings.HasSuffix(file.Name(), ".zip") {
+					log.Info("Attempting to unzip " + file.Name() + " before proceeding to upload")
+					path, _ = utils.ExtractFile(filepath.Join(path, file.Name()), "dsym")
```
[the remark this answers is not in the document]
jasonMcCullough: Yep for sure we'll need that, I can try and see if we can improve that so that it's self-contained in the util function

pkg/upload/dsym.go
```diff
@@ -0,0 +1,176 @@
...
+			}
+
+			// Build the dsymPath from build settings
+			dsymPath = filepath.Join(buildSettings.ConfigurationBuildDir, buildSettings.DsymName, "Contents", "Resources", "DWARF")
+
+			// Check if dsymPath exists, if not, try alternative path instead
+			err = utils.Path(dsymPath).Validate()
+			if err != nil {
+				log.Info("Could not find dSYM in expected location: " + utils.DisplayBlankIfEmpty(dsymPath))
+				dsymPath = filepath.Join(buildSettings.ConfigurationBuildDir, strings.TrimSuffix(buildSettings.DsymName, ".dSYM"))
```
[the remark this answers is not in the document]
jasonMcCullough: Having looked into this more I've discovered some of this is not needed. The artifact in `.app` is an iOS application bundle apparently, which when used with dwarfdump gave a valid UUID which made me think it could be used as an alternative if `.app.dSYM` is not present, but I don't think that is possible and can probably just error if it is missing.

pkg/upload/dsym.go
```diff
@@ -0,0 +1,176 @@
...
+
+				err = utils.Path(dsymPath).Validate()
+				if err != nil {
+					return err
```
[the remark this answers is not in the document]
jasonMcCullough: We could log something like "Try re-building your project or explicitly define the location of your dSYM with --dsym-path"?

pkg/upload/dsym.go
```diff
@@ -0,0 +1,176 @@
...
+	dryRun bool,
+) error {
+
+	var buildSettings *ios.XcodeBuildSettings
+	var dsyms *[]*ios.DsymFile
+	var plistData *ios.PlistData
+	var uploadOptions map[string]string
+
+	for _, path := range paths {
+		uploadInfo, err := ios.ProcessPathValue(path, projectRoot)
```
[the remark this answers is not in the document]
jasonMcCullough: The `projectRoot` is passed in just to make sure we use it instead of `path` if `path` is a directory. When `--project-root` takes precedence, we let the user know "--project-root flag set, it's value takes precedence and will be used for upload" in case they don't know why the project root was different to what they expected.

pkg/upload/dsym.go
jasonMcCullough: I've moved all unzip logic to `GetDsymsForUpload` now and I agree this is confusing. If an unsucessful unzip happens we log a warn level log "Could not unzip " + file.Name() + " to a temporary directory, skipping", try to remove any temporary directory that may have gotten created and then keep going with the other uploads.

pkg/upload/dsym.go
```diff
@@ -0,0 +1,176 @@
...
+
+				// Check if the variables are empty, set if they are abd log that we are using setting from the plist file
+				if versionName == "" {
+					versionName = plistData.VersionName
+					log.Info("Using version name from Info.plist: " + utils.DisplayBlankIfEmpty(versionName))
+
+				}
+
+				if apiKey == "" {
+					apiKey = plistData.BugsnagProjectDetails.ApiKey
```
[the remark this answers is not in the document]
jasonMcCullough: I was trying to think of a way that would let the user know that we had no choice but to fetch a blank api-key value from the plist (since it was not defined explicitly by them), so that when the upload server complains about a missing api key they know to go and add it if they were expecting it to be there. Is there a limitation to how useful that might be? I'm aware it creates a little more noise though too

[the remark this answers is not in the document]
tomlongridge: It will trigger the validation in `BuildDsymUploadOptions` to say that API key is missing – for consistency with the other platforms (e.g. [here](https://github.com/bugsnag/bugsnag-cli/blob/54914517e7803c0f6d2e6f43bbf28b248b2d0205/pkg/upload/android-proguard.go#L118)), I'd say only log if it's non-empty.

pkg/ios/dwarfdump-utils.go
```diff
@@ -0,0 +1,82 @@
...
+// GetDsymsForUpload returns information on the valid dSYM files found in a given path
+func GetDsymsForUpload(path string) (*[]*DsymFile, error) {
+	filesFound, _ := os.ReadDir(path)
+	var dsymFiles []*DsymFile
+
+	switch len(filesFound) {
+	case 0:
+		return nil, errors.Errorf("No dSYM files found in expected location '%s'", path)
+	default:
+		if isDwarfDumpInstalled() {
```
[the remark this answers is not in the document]
tomlongridge: Not a biggie, but I guess I was thinking that if you check for dwarfdump being installed and then for >= 1 dsym, exiting accordingly right at the start, the loop doesn't need to be so nested and it's easier to read.

I guess that's a style thing: some people don't like multiple return statements.

Happy to go with either approach.

pkg/upload/dsym.go
```diff
@@ -0,0 +1,160 @@
...
+			// Only when the dsym path is not set, try and work out the scheme
+			if dsymPath == "" {
+				// If the scheme is not set explicitly, try to find it
+				scheme, err = ios.GetDefaultScheme(path, projectRoot)
+				if err != nil {
+					return err
+				}
+			}
+		}
+		log.Info("Using scheme: " + utils.DisplayBlankIfEmpty(scheme))
```
tomlongridge, line 77: Move this into the proceeding if-statement as the scheme isn't being "used" if there is a dsym path already.

pkg/upload/dsym.go
tomlongridge: I see – given we'd usually expect just one file I feel like this could be an `Error` message and focussing a bit less on the temp dir (as this isn't something the user knows or can control). The "skipping" bit combined with a warning makes it all seem optional. 

How about:

```
log.Error("Could not unzip " + file.Name() + " for upload")
```

I guess we'd have to update the `log.Error` method to allow a non-exiting error. This makes me think it should honour `--fail-on-upload`, but maybe that's stretching the option too far?

pkg/upload/dsym.go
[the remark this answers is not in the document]
tomlongridge: I'm not quite clear why it's being used later on in the Xcode project file parsing?

pkg/ios/dwarfdump-utils.go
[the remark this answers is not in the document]
jasonMcCullough: Oh yeah that's fine by me, I think that would read better too.

pkg/upload/dsym.go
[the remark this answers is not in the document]
jasonMcCullough: Yeah this should not be the case which makes sense since the path is going to have the option of containing a "project root" so I have now removed it to reflect this change.

pkg/upload/dsym.go
jasonMcCullough: I completely agree with the above, and as discussed this flag along with the --ignore-missing-dwarf and --ignore-empty-dsym will be implemented as part of my other PR.

pkg/upload/dsym.go
```diff
@@ -0,0 +1,165 @@
...
+		// If dsymPath is not set explicitly, use uploadInfo to set it (if available)
+		// uploadInfo.DsymPath is set when <path> is recognised as a dsym path
+		if dsymPath == "" && uploadInfo.DsymPath != "" {
+			dsymPath = uploadInfo.DsymPath
+		}
+
+		// If projectRoot is not set explicitly, use uploadInfo to set it
+		if projectRoot == "" {
+			projectRoot = uploadInfo.ProjectRoot
+		}
```
tomlongridge, line 58: It feels to me that this section is a little over-complicated with the intermediate struct – is there any reason not to have separate functions to set the project root (if not provided) and dSYM path (if not provided); or use multiple return values from `ProcessPathValue`? (However to me, it feels like `ProcessPathValue` is trying to do too much and making it harder to read in doing so.)

jasonMcCullough: I'm in agreement that `ProcessPathValue` is a little "busy" in it's current form. I'll see what I can do to split things out a bit better and maybe pull out some of these empty checks into an intermediate validation function or something perhaps?

tomlongridge: IMO there doesn't seem much point in the `DsymUploadInfo` struct now that we've established that project root isn't going to be used for any other input and we're not defaulting it to cwd. My suggestion would be to have 2 functions and handle the override, validation and default values for `projectRoot` and `path` separately.

test/testdata/ios/dsym-test-fixtures/fl-project/fastlane/Fastfile
tomlongridge, line 1: Is this file actually in use? I wasn't expecting Fastlane integration to be working yet.

jasonMcCullough: No this isn't in use yet, I'll just remove it until we need it

PR #78, opened 2024-02-13
Ability to perform multiple dSYM uploads from one path

jasonMcCullough:
## Goal

To better coincide with our soon to be legacy dSYM upload script. ([see here](https://github.com/bugsnag/bugsnag-dsym-upload/commit/ea3fa6f684f445eff5ee696374eb4b7c391e8948))

## Design

I made some adjustments to the behaviour so that we can:
1) Take one path, either as an explicit `--dsym-path` argument or from `<path>`
2) Check if it contains dSYMs, if it does, use those dSYM paths for uploading one or more dSYMs
3) Also proceeds to using `<path>` as project root if it is a directory with no dSYMs

## Changeset

Few changes made to xcode utility functions and to the "core" functionality defined within `ProcessDsym`.

## Testing

Manually tested/debugged locally and will be covered by Maze Runner testing which is currently in the works in the upstream branch for this.


joshedney (approved):
Lgtm

jasonMcCullough:
@joshedney I made some adjustments yesterday after your initial review so bare with me once I get those adjustments tested 🙏

joshedney (approved):
LGTM!

PR #79, opened 2024-02-14
Overhaul to how dSYM uploads are validated and processed

jasonMcCullough:
## Goal

To make the dSYM upload logic a bit easier to read and add some more flexibility.

Also includes edge case where `<path>` is pointing to an xcode project or workspace but the user has not built their dSYMs in the default location so can define it with `--dsym-path`.

## Design

A bit more flexibility between `--dsym-path` and `<path>` usage.

## Changeset

`--dsym-path`
* Can be a directory containing one of more dSYMs
* Can be a .zip containing one or more dSYMs
* Can be a path to one dSYM

`<path>`
* Can be a directory containing one of more dSYMs
* Can be a .zip containing one or more dSYMs
* Can be a path to one dSYM

## Testing

Integration Tests with Maze Runner to follow


pkg/upload/dsym.go
```diff
@@ -36,82 +38,105 @@ func ProcessDsym(
...
-		if scheme != "" && len(uploadInfo.DsymPaths) == 0 {
-			_, err = ios.IsSchemeInPath(path, scheme)
-			if err != nil {
-				return err
-			}
-			log.Info("Using scheme: " + scheme)
-		} else {
+		if ios.IsPathAnXcodeProjectOrWorkspace(path) {
+			projectRoot = ios.GetDefaultProjectRoot(path, projectRoot)
+			log.Info("Using '" + projectRoot + "' as project root, this can be changed at any time with a --project-root flag")
```
tomlongridge, line 63: Suggest toning it down a bit to avoid causing alarm, as this is perfectly valid:

```suggestion
			log.Info("Defaulting to '" + projectRoot + "' as the project root")
```

pkg/ios/dsym-utils.go
```diff
@@ -0,0 +1,138 @@
...
+
+			fileName := filepath.Base(path)
+			log.Info("Attempting to unzip " + fileName + " before proceeding to upload")
+
+			var err error
+			tempDir, err = utils.ExtractFile(path, "dsym")
+
+			if err != nil {
+				// TODO: This will be downgraded to a warning with --fail-on-upload in near future
+				log.Error("Could not unzip "+fileName+" to a temporary directory, skipping", 1)
```
lemnik, line 46: nit: formatting

```suggestion
				log.Error("Could not unzip " + fileName + " to a temporary directory, skipping", 1)
```

pkg/upload/dsym.go
```diff
@@ -36,82 +38,105 @@ func ProcessDsym(
...
-				log.Info("Using dSYM path: " + dsymPath)
+			if dsymPath == "" {
+				// Build the dsymPath from build settings
+				// Which is built up to look like: /Users/Path/To/Config/Build/Dir/MyApp.app.dSYM
+				dsymPath = filepath.Join(buildSettings.ConfigurationBuildDir, buildSettings.DsymName)
+
+				// Check if dsymPath exists before proceeding
+				if utils.Path(dsymPath).Validate() != nil {
+					// TODO: This will be toggled between Error and Warn with --ignore-missing-dwarf in near future
+					log.Error("Could not find dSYM with scheme '"+scheme+"' in expected location: "+utils.DisplayBlankIfEmpty(dsymPath)+"\n\n"+
```
lemnik, line 98: nit: formatting

```suggestion
					log.Error("Could not find dSYM with scheme '" + scheme + "' in expected location: " + utils.DisplayBlankIfEmpty(dsymPath) + "\n\n" +
```

pkg/ios/dsym-utils.go
```diff
@@ -0,0 +1,138 @@
...
+			} else {
+				log.Info("Unzipped " + fileName + " to " + tempDir + " for uploading")
+				dsymLocations = findDsyms(tempDir)
+
+			}
+
+		} else if strings.HasSuffix(path, ".dSYM") {
```
lemnik, line 53: General thought: is it worth considering case-insensitive checks when doing these? Especially given how inconsistent things get between platforms?

pkg/upload/dsym.go
```diff
@@ -36,82 +38,105 @@ func ProcessDsym(
...
 				}
-				log.Info("Using scheme: " + scheme)
+
 			}
-		}
 
-		// If the dsymPath is not fed in via <path>
-		if dsymPath == "" && len(uploadInfo.DsymPaths) == 0 {
+			var err error
 			buildSettings, err = ios.GetXcodeBuildSettings(path, scheme)
```
tomlongridge, line 85: scheme can be nil here - if we didn't find it in the default. So this should be conditioned and on L90 it should check `buildSettings` being nil.

pkg/upload/dsym.go
```diff
@@ -36,82 +38,105 @@ func ProcessDsym(
...
+			dsymPath = path
 		}
 
-		dsyms, err = ios.GetDsymsForUpload(uploadInfo.DsymPaths)
-		if err != nil {
-			return err
+		if dsymPath != "" {
+			var tempDir string
+			dwarfInfo, tempDir, _ = ios.FindDsymsInPath(dsymPath)
+			if len(dwarfInfo) > 0 && projectRoot == "" {
```
tomlongridge, line 113: This might read a bit better as an else-if on L121 (after the zero length check)

pkg/ios/dsym-utils.go
tomlongridge: .zip more likely than .dSYM for sure, but calling toLower for all these cases seems like a good shout.

tomlongridge (approved):

tomlongridge (approved):

PR #80, opened 2024-02-19
[PLAT-11498] Maze runner setup, integration tests and test-driven changes

jasonMcCullough:
## Goal

To adequately integration test the new dSYM upload functionality.

## Design

Based on [pre-existing](https://github.com/bugsnag/bugsnag-dsym-upload/tree/master/features) test cases that were used to measure the success of the previous dSYM upload method.

## Changeset

* A version of Maze Runner that supports requests to `/dsym`
* Various feature files containing various use case scenarios.

## Testing

Maze Runner


pkg/ios/dsym-utils.go
```diff
@@ -47,11 +47,10 @@ func FindDsymsInPath(path string) ([]*DwarfInfo, string, error) {
 			} else {
 				log.Info("Unzipped " + fileName + " to " + tempDir + " for uploading")
 				dsymLocations = findDsyms(tempDir)
-
 			}
 
-		} else if strings.HasSuffix(strings.ToLower(path), ".dsym") {
-			// If path points to a .dSYM file, then we will use it as is
+		} else {
+			// If path points to a file, then we will assume it is a dSYM and use it as-is
```
jasonMcCullough, line 53: I changed this because I thought of an edge case where the path could be a file, but not have the `.dSYM` suffix i.e. the actual file that resides in the `DWARF` directory of a path. So if it is referenced directly, we can still handle it.

pkg/ios/dsym-utils.go
```diff
@@ -64,7 +63,16 @@ func FindDsymsInPath(path string) ([]*DwarfInfo, string, error) {
 		}
 
 		for _, dsymLocation := range dsymLocations {
-			filesFound, _ := os.ReadDir(dsymLocation)
+			filesFound, err := os.ReadDir(dsymLocation)
+
+			if err != nil {
```
jasonMcCullough, line 68: This coincides with my previous comment

PR #81, opened 2024-02-20
[PLAT-11497] Implement --fail-on-upload, --ignore-missing-dwarf and --ignore-empty-dsym flags

jasonMcCullough:
## Goal

To downgrade errors to warning which would allow valid uploads to proceed out of batches of uploads that may contain invalid uploads etc.

## Design

Follows the same approach as our previous solution (bugsnag-dsym-upload).

## Changeset

Some conditions dotted around where necessary to either error out with an error level log or warn and continue.

## Testing

Some manual testing done locally and a couple of scenarios tested with Maze Runner.


pkg/upload/dsym.go
```diff
@@ -97,9 +103,13 @@ func ProcessDsym(
...
-						// TODO: This will be toggled between Error and Warn with --ignore-missing-dwarf in near future
-						log.Error("Could not find dSYM with scheme '"+scheme+"' in expected location: "+utils.DisplayBlankIfEmpty(dsymPath)+"\n\n"+
-							"Check that the scheme correlates to the above dSYM location, try re-building your project or specify the dSYM path using --dsym-path\n", 1)
+						if ignoreMissingDwarf {
+							log.Warn("Could not find dSYM with scheme '" + scheme + "' in expected location: " + utils.DisplayBlankIfEmpty(dsymPath) + "\n\n" +
+								"Check that the scheme correlates to the above dSYM location, try re-building your project or specify the dSYM path using --dsym-path\n")
+						} else {
+							log.Error("Could not find dSYM with scheme '"+scheme+"' in expected location: "+utils.DisplayBlankIfEmpty(dsymPath)+"\n\n"+
+								"Check that the scheme correlates to the above dSYM location, try re-building your project or specify the dSYM path using --dsym-path\n", 1)
+						}
```
tomlongridge, line 112: On reflection, I don't think this is the scenario we're talking about for missing dwarf information.

I think we'd be OK just to validate the path and set it if it exists, otherwise don't set `dsymPath` and let it error later on.

pkg/ios/dsym-utils.go
```diff
@@ -133,7 +149,8 @@ func findDsyms(root string) []string {
 			return err
 		}
 
-		if strings.HasSuffix(strings.ToLower(info.Name()), ".dsym") {
+		// If the file is a dSYM, add it to the list (unless it resides within the __MACOSX directory)
+		if strings.HasSuffix(strings.ToLower(info.Name()), ".dsym") && !strings.Contains(strings.ToLower(path), "__macosx") {
```
jasonMcCullough, line 153: I changed this because we used to do the same thing in bugsnag-dsym-upload ([see here](https://github.com/bugsnag/bugsnag-dsym-upload/blob/47fee06af3a6d92e8aafd3c9d3453649ded03e13/bin/bugsnag-dsym-upload#L136)).

pkg/ios/dsym-utils.go
```diff
@@ -42,8 +42,11 @@ func FindDsymsInPath(path string) ([]*DwarfInfo, string, error) {
 			tempDir, err = utils.ExtractFile(path, "dsym")
 
 			if err != nil {
-				// TODO: This will be downgraded to a warning with --fail-on-upload in near future
-				log.Error("Could not unzip "+fileName+" to a temporary directory, skipping", 1)
+				if failOnUpload {
+					log.Warn("Could not unzip " + fileName + " to a temporary directory, skipping")
+				} else {
+					log.Error("Could not unzip "+fileName+" to a temporary directory, skipping", 1)
```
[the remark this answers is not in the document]
jasonMcCullough: I'm using GoLand's default formatting which forces this squashing for some reason

tomlongridge (approved):

PR #83, opened 2024-03-01
[PLAT-11638] Add fallback endpoint for DSYM uploads

joshedney:
## Goal

- Add fallback endpoint for DSYM uploads when processing the DSYM API request.

## Changeset

Following the design for Android Proguard, we're going to attempt to send the request to the `\dsym` endpoint. However, if that returns a 404, we will attempt to send the request to the fallback URL.

## Testing

Covered by CI

### DIFF
--- a/pkg/ios/dsym-utils.go
+++ b/pkg/ios/dsym-utils.go
@@ -42,7 +42,7 @@ func FindDsymsInPath(path string, ignoreEmptyDsym, ignoreMissingDwarf bool) ([]*
 			tempDir, err = utils.ExtractFile(path, "dsym")
 
 			if err != nil {
-				log.Error("Could not unzip " + fileName + " to a temporary directory, skipping", 1)
+				log.Error("Could not unzip "+fileName+" to a temporary directory, skipping", 1)
 
 			} else {
 				log.Info("Unzipped " + fileName + " to " + tempDir + " for uploading")
@@ -91,7 +91,7 @@ func FindDsymsInPath(path string, ignoreEmptyDsym, ignoreMissingDwarf bool) ([]*
 					if ignoreEmptyDsym {
 						log.Warn("Skipping empty file: " + file.Name())
 					} else {
-						log.Error("Skipping empty file: " + file.Name(), 0)
+						log.Error("Skipping empty file: "+file.Name(), 0)
 					}
 				}
 			}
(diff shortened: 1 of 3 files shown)


pkg/ios/dsym-utils.go
```diff
@@ -42,7 +42,7 @@ func FindDsymsInPath(path string, ignoreEmptyDsym, ignoreMissingDwarf bool) ([]*
 			tempDir, err = utils.ExtractFile(path, "dsym")
 
 			if err != nil {
-				log.Error("Could not unzip " + fileName + " to a temporary directory, skipping", 1)
+				log.Error("Could not unzip "+fileName+" to a temporary directory, skipping", 1)
```
tomlongridge, line 45: Have we got some local linter settings conflicting with each other? See [previous review comment](https://github.com/bugsnag/bugsnag-cli/pull/79#discussion_r1492609086).

I think you can remove this change from the PR, but it would be good if you can look into checking-in some project linting rules.

joshedney: You're right, for some reason my IDE was using a different linter to the one we have in the `makefile`

pkg/upload/dsym.go
```diff
@@ -187,15 +188,17 @@ func ProcessDsym(
 			err = server.ProcessFileRequest(endpoint+"/dsym", uploadOptions, fileFieldData, timeout, retries, dsym.UUID, dryRun)
 
 			if err != nil {
-				if failOnUpload {
-					return err
-				} else {
-					log.Warn(err.Error())
+				if strings.Contains(err.Error(), "404 Not Found") {
+					log.Info("Trying " + endpoint)
```
jasonMcCullough, line 192: Should be elaborate for the user what happened here before the retry? cc: @tomlongridge for your opinion too

pkg/upload/dsym.go
```diff
@@ -187,15 +188,17 @@ func ProcessDsym(
 			err = server.ProcessFileRequest(endpoint+"/dsym", uploadOptions, fileFieldData, timeout, retries, dsym.UUID, dryRun)
 
 			if err != nil {
-				if failOnUpload {
```
jasonMcCullough, line 190: We can't downgrade to WARN without this check, so the integration test is failing because of this

pkg/upload/dsym.go
tomlongridge: The user doesn't have any control over which path to use, so I'd actually suggest we don't log at all here. But I agree, we should either log both attempts or none. (Consider what we do on other platforms – maybe introducing a log the endpoint used across all platforms is something we raise as a small future task.)

PR #85, opened 2024-03-05
[PLAT-11668] Remove `--dsym-path` option and add `--xcode-project` option

joshedney:
## Goal

To match how our other commands work we're going to remove the option `--dsym-path` and set it so that the default path argument can be either an Xcode project, single dSYM file or a dSYM zip file.

## Changeset

Adjust how the dSYM upload functionality processes files passed to the `path` argument to determine if we're dealing with an `xcodeproj/workspace` file or if we're dealing with one of the dSYM formats before we attempt to process the file.

## Testing

Covered by CI

### DIFF
--- a/pkg/upload/dsym.go
+++ b/pkg/upload/dsym.go
@@ -72,14 +72,14 @@ func ProcessDsym(
 		}
 
 		if xcodeProjPath != "" {
-			projectRoot = ios.GetDefaultProjectRoot(path, projectRoot)
+			projectRoot = ios.GetDefaultProjectRoot(xcodeProjPath, projectRoot)
 			log.Info("Defaulting to '" + projectRoot + "' as the project root")
 
 			// Get build settings and dsymPath
 
 			// If scheme is set explicitly, check if it exists
 			if scheme != "" {
-				_, err := ios.IsSchemeInPath(path, scheme)
+				_, err := ios.IsSchemeInPath(xcodeProjPath, scheme)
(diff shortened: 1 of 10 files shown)


pkg/upload/dsym.go
```diff
@@ -27,10 +27,8 @@ type Dsym struct {
 
 func ProcessDsym(
 	apiKey string,
-	versionName string,
 	scheme string,
-	dev bool,
-	dsymPath string,
+	xcodeProjPath string,
```
tomlongridge, line 31: `overwrite` isn't a parameter in the API, so should be removed.

pkg/upload/dsym.go
```diff
@@ -111,9 +110,26 @@ func ProcessDsym(
...
+
+					if scheme != "" {
+						var err error
+						buildSettings, err = ios.GetXcodeBuildSettings(xcodeProjPath, scheme)
+						if err != nil {
+							return err
+						}
+					}
+				}
+			}
```
tomlongridge, line 132: To avoid this duplication, could we do the following before the if-statement on L67:

```
if ios.IsPathAnXcodeProjectOrWorkspace(path) {
    if xcodeProjPath == ""  {
        xcodeProjPath = path
    }
} else {
    dsymPath = path
}

if xcodeProjPath != "" {
    // As above
} 
```

tomlongridge (approved):

PR #90, opened 2024-03-13
[PLAT-11746] Deprecate `--fail-on-upload-error` and handling for duplicate file response

joshedney:
## Goal

Deprecate the `--fail-on-upload-error` flag and log a warning if the flag is used. From now on when an upload fails it will always fail unless the API server returns a `409` error where it will instead log a warning informing the end user  that the duplicate file has been skipped.

I have also updated the formatting of some of the longer functions to fall inline with recent styling changes to make them easier to read.

## Testing

Covered by CI

### DIFF
--- a/main.go
+++ b/main.go
@@ -43,7 +43,7 @@ func main() {
 	}
 
 	if commands.FailOnUploadError {
-		log.Warn("The `--fail-on-upload-error` flag is deprecated and will be removed in a future release.")
+		log.Warn("The `--fail-on-upload-error` flag is deprecated and will be removed in a future release. All commands now fail if the upload is unsuccessful.")
 	}
 
 	switch ctx.Command() {
--- a/main.go
+++ b/main.go
@@ -42,6 +42,10 @@ func main() {
 		log.Info("Performing dry run - no data will be sent to BugSnag")
 	}
 
+	if commands.FailOnUploadError {
+		log.Warn("The `--fail-on-upload-error` flag is deprecated and will be removed in a future release.")
+	}
+
 	switch ctx.Command() {
 
 	case "upload all <path>":
@@ -58,7 +62,6 @@ func main() {
 			commands.Upload.Retries,
 			commands.Upload.Overwrite,
 			commands.ApiKey,
-			commands.FailOnUploadError,
 			commands.DryRun,
 		)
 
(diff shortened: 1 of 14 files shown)


main.go
```diff
@@ -42,6 +42,10 @@ func main() {
 		log.Info("Performing dry run - no data will be sent to BugSnag")
 	}
 
+	if commands.FailOnUploadError {
+		log.Warn("The `--fail-on-upload-error` flag is deprecated and will be removed in a future release.")
```
tomlongridge, line 46:
```suggestion
		log.Warn("The `--fail-on-upload-error` flag is deprecated and will be removed in a future release. All commands now fail if the upload is unsuccessful.")
```

pkg/android/process-uploads.go
```diff
@@ -29,12 +42,8 @@ func UploadAndroidNdk(fileList []string, apiKey string, applicationId string, ve
 		err = server.ProcessFileRequest(endpoint+"/ndk-symbol", uploadOptions, fileFieldData, timeout, retries, file, dryRun)
 
 		if err != nil {
-			if numberOfFiles > 1 {
-				if failOnUploadError {
-					return err
-				} else {
-					log.Warn(err.Error())
-				}
+			if strings.Contains(err.Error(), "409") && strings.Contains(err.Error(), "duplicate") {
```
tomlongridge, line 45: Do you have any reason to believe it could return a 409 for any other reason? If not, I think a check for status code alone would be sufficient.

pkg/android/process-uploads.go
```diff
@@ -29,12 +42,8 @@ func UploadAndroidNdk(fileList []string, apiKey string, applicationId string, ve
 		err = server.ProcessFileRequest(endpoint+"/ndk-symbol", uploadOptions, fileFieldData, timeout, retries, file, dryRun)
```
tomlongridge, line 42: How about `ProcessFileRequest` handling this internally and not returning an `err`? It could log the message (at info level) to say it's skipped the file; and also the successful "Uploaded" message so the caller doesn't need to do this.

@lemnik – thoughts on this?

I seem to recall I've previously suggested the caller logs the success message as it has the context about what is being uploaded (e.g. "dSYM uploaded"), but actually I don't see any harm in us just using the file name like we have here for all the uploads.

pkg/android/process-uploads.go
joshedney: I do not but I thought the added check would ensure that we are sure its a duplicate.

tomlongridge: The intent of 409 being to mark the file as duplicate is clear – I'd rather this just checked the code and not introduce potential future problems if the less-defined message ever changes.

tomlongridge (approved):

PR #91, opened 2024-03-22
[PLAT-11829] Ensure that `timeout` is sent to the correct options during upload

joshedney:
## Goal

Ensure that the uploads are retried the correct number of times for Unity Android

## Changeset

The value for timeout (`300`) was being passed to the retries value for the unity android upload resulting in 300 retires.

## Testing

Covered by CI


lemnik (approved):
LGTM

PR #92, opened 2024-03-22
Fix dex-derived BuildUUID for proguard uploads

lemnik:
## Goal
Attempt to automatically derive a buildUuid from the `classes.dex` files for `mapping.txt` files.

## Design
If the `buildUuid` for a proguard file is not specified as either an option, or in the `AndroidManifest.xml` then the tool will attempt to locate the appropriate `classesN.dex` files for the `mapping.txt` file, and derive the BuildUUID from these.

Added a new `--no-build-uuid` (which is mutually exclusive with `--build-uuid`) to the various Android upload options. If this is specified the `buildUuid` is set to `""` and no attempt is made to find or generate a BuildUUID.

## Testing
Manually tested


pkg/android/aab.go
```diff
@@ -58,7 +58,7 @@ func MergeUploadOptionsFromAabManifest(path string, apiKey string, applicationId
 					log.Info("Using " + aabUploadOptions["buildUuid"] + " as build ID from dex signatures")
 				}
 			}
-		} else if aabUploadOptions["buildUuid"] == "none" {
+		} else if aabUploadOptions["buildUuid"] == "none" || noBuildUuid {
```
tomlongridge, line 61: Shall we ticket and/or comment here to remove this in a future release?

tomlongridge (approved):
LGTM - with a changelog

PR #94, opened 2024-04-11
Update supported-platforms.yml

goblinbr:
Fixes #93

Issue #93, opened 2024-04-11
Platform with type "Windows_NT" and architecture "x64" is not supported by @bugsnag/cli


goblinbr:
on `npm install --save-dev @bugsnag/cli` it's always showing this error:
```
npm ERR! code 1
npm ERR! path C:\projetos\AplicativoAgronomos\node_modules\@bugsnag\cli
npm ERR! command failed
npm ERR! command C:\WINDOWS\system32\cmd.exe /d /s /c node install.js
npm ERR! Platform with type "Windows_NT" and architecture "x64" is not supported by @bugsnag/cli.
npm ERR! Your system must be one of the following:
npm ERR!
npm ERR! [
npm ERR!   {
npm ERR!     "Type": "Windows",
npm ERR!     "Architecture": "x64",
npm ERR!     "Artifact": "x86_64-windows-bugsnag-cli.exe"
npm ERR!   },
npm ERR!   {
npm ERR!     "Type": "Windows",
npm ERR!     "Architecture": "i386",
npm ERR!     "Artifact": "i386-windows-bugsnag-cli.exe"
npm ERR!   },
npm ERR!   {
npm ERR!     "Type": "Linux",
npm ERR!     "Architecture": "x64",
npm ERR!     "Artifact": "x86_64-linux-bugsnag-cli"
npm ERR!   },
npm ERR!   {
npm ERR!     "Type": "Linux",
npm ERR!     "Architecture": "i386",
npm ERR!     "Artifact": "i386-linux-bugsnag-cli"
npm ERR!   },
npm ERR!   {
npm ERR!     "Type": "Darwin",
npm ERR!     "Architecture": "x64",
npm ERR!     "Artifact": "x86_64-macos-bugsnag-cli"
npm ERR!   },
npm ERR!   {
npm ERR!     "Type": "Darwin",
npm ERR!     "Architecture": "arm64",
npm ERR!     "Artifact": "arm64-macos-bugsnag-cli"
npm ERR!   }
npm ERR! ]

npm ERR! A complete log of this run can be found in: C:\Users\Rodrigo\AppData\Local\npm-cache\_logs\2024-04-11T12_30_34_219Z-debug-0.log
```

Node version: v18.20.1
NPM version: 10.5.2
OS: Windows 11 (23H2 22631.3447)

[...]

PR #95, opened 2024-04-15
[PLAT-11977] Include Windows_NT in supported-platforms.yml

jasonMcCullough:
@goblinbr raised [an issue](https://github.com/bugsnag/bugsnag-cli/issues/93) that highlighted the need for including `Windows_NT` in our supported platforms config file. 

Their original fix can be viewed [here](https://github.com/bugsnag/bugsnag-cli/pull/94) which includes 32 and 64 bit definitions identically defined as our pre-existing `Windows` definitions.


joshedney (approved):
LGTM!

PR #98, opened 2024-04-18
Allow more flexible path searching when uploading NDK symbol files

hannah-smartbear:
- Added a function which searches for merged_native_libs directory based on an expected path
- The function checks:
android/app/build/intermediates/merged_native_libs
app/build/intermediates/merged_native_libs
build/intermediates/merged_native_libs
intermediates/merged_native_libs
merged_native_libs
- It also checks back through the path in case the command was run from somewhere within the merged_native_libs directory
- This Jira ticket originally came from wanting to be able to use the CLI tool to upload NDK symbol files from an Android library and this should achieve that
- I also tested my changes locally
- I added some test scenarios to android-ndk.feature


pkg/upload/android-ndk.go
```diff
@@ -227,3 +210,36 @@ func ProcessAndroidNDK(
 
 	return nil
 }
+
+func findNativeLibPath(arr []string, path string) (string, error) {
```
joshedney, line 214: Could you move this to `pkg/android/` with the other android related tooling

joshedney (approved):
LGTM!

PR #99, opened 2024-04-23
[PLAT-11999] Verbose logging

joshedney:
## Goal

Update the logging for the Bugsnag CLI to support additional log levels.

Integrate a new log wrapper that will allow us to drop different logging libraries in without much reconfiguration throughout the CLI.

Move over to using `logrus` for logging

When the `--verbose` and `--dry-run` flag are passed, the CLI prints out the payload to be sent to the API, minus the file data.

e.g.

```
[INFO] (dryrun) Skipping upload of libunity.so to https://upload.bugsnag.com:443/ndk-symbol
[INFO] (dryrun) Upload payload:
{
    "apiKey": "APIKEY",
    "appId": "com.DefaultCompany.Myproject3",
    "sharedObjectName": "libunity.so",
    "versionCode": "1",
    "versionName": "0.1"
}
```
Move information messages down to the debug log level so that they only print when passing the `--verbose` flag.

## Testing

Covered by CI


pkg/log/logrus-wrapper.go
```diff
@@ -0,0 +1,98 @@
...
+
+func (l *LogrusLogger) addContextCommonFields(fields map[string]interface{}) {
+	if l.ctx != nil {
+		for k, v := range l.ctx.Value("commonFields").(map[string]interface{}) {
+			if _, ok := fields[k]; !ok {
+				fields[k] = v
+			}
+		}
+	}
+}
```
tomlongridge, line 98: Is this used?

pkg/server/request.go
```diff
@@ -83,28 +83,31 @@ func buildFileRequest(url string, fieldData map[string]string, fileFieldData map
...
 		if err != nil {
 			if strings.Contains(err.Error(), "409") {
-				log.Warn(fmt.Sprintf("Duplicate file detected, skipping upload of %s", filepath.Base(fileName)))
+				logger.Warn(fmt.Sprintf("Duplicate file detected, skipping upload of %s", filepath.Base(fileName)))
 			} else {
 				return err
 			}
 		} else {
-			log.Success(fmt.Sprintf("Uploaded %s", filepath.Base(fileName)))
+			logger.Info("Uploaded " + filepath.Base(fileName))
```
tomlongridge, line 104: Deliberate removal of sprintf?

pkg/upload/all.go
```diff
@@ -20,18 +20,19 @@ func All(
...
 
 	fileList, err := utils.BuildFileList(paths)
 
 	if err != nil {
-		log.Error(" error building file list", 1)
+		logger.Fatal(" error building file list")
 	}
 
-	log.Info("File list built..")
+	logger.Info("File list built..")
```
tomlongridge, line 35: Feels like the two `Info` log calls are unnecessary, when comparing this to other commands. Just an error message if it fails seems sufficient.

Likewise with the Dart command.

pkg/upload/all.go
```diff
@@ -20,18 +20,19 @@ func All(
...
 
 	// Build the file list from the path(s)
-	log.Info("building file list...")
+	logger.Info("building file list...")
 
 	fileList, err := utils.BuildFileList(paths)
 
 	if err != nil {
-		log.Error(" error building file list", 1)
+		logger.Fatal(" error building file list")
```
tomlongridge, line 32: We should also include the error message

```suggestion
		logger.Fatal("Error building file list")
```

pkg/upload/dart.go
```diff
@@ -125,7 +126,7 @@ func Dart(
 
 			continue
 		}
-		log.Info(fmt.Sprintf("Skipping %s", file))
+		logger.Debug(fmt.Sprintf("Skipping %s", file))
```
tomlongridge, line 129:
```suggestion
		logger.Debug(fmt.Sprintf("Skipping %s - unsupported platform", file))
```

pkg/upload/dsym.go
```diff
@@ -146,14 +147,14 @@ func ProcessDsym(
...
-					log.Info(fmt.Sprintf("Using API key from Info.plist: %s", apiKey))
+					logger.Debug(fmt.Sprintf("Using API key from Info.plist: %s", apiKey))
 				}
 			}
 		}
 
 		for _, dsym := range dwarfInfo {
 			dsymInfo := fmt.Sprintf("(UUID: %s, Name: %s, Arch: %s)", dsym.UUID, dsym.Name, dsym.Arch)
-			log.Info(fmt.Sprintf("Uploading dSYM %s", dsymInfo))
+			logger.Info(fmt.Sprintf("Uploading dSYM %s", dsymInfo))
```
tomlongridge, line 157: Given the "Uploading ..." message is emitted from `ProcessFileRequest`, perhaps make this debug and the message a bit different?

```suggestion
			logger.Debug(fmt.Sprintf("Processing dSYM %s", dsymInfo))
```

pkg/server/request.go
```diff
@@ -83,28 +83,31 @@ func buildFileRequest(url string, fieldData map[string]string, fileFieldData map
...
 		}
 	} else {
-		log.Info(fmt.Sprintf("(dryrun) Skipping upload of %s to %s", filepath.Base(fileName), endpoint))
+		logger.Info(fmt.Sprintf("(dryrun) Skipping upload of %s to %s", filepath.Base(fileName), endpoint))
+		logger.Info("(dryrun) Upload payload:")
+		prettyUploadOptions, _ := utils.PrettyPrintMap(uploadOptions)
```
tomlongridge, line 109: > When the --verbose and --dry-run flag are passed, the CLI prints out the payload to be sent to the API, minus the file data.

Does this correctly omit the file data as described?

pkg/server/request.go
joshedney: Restored sprintf use

pkg/server/request.go
joshedney: Yes, `uploadOptions` doesn't contain the file data, but does include the name of the file that is being uploaded. The file data is stored within `fileFieldData`

tomlongridge (approved):

PR #102, opened 2024-05-15
Using switches to restrict input for source control provider option

hannah-smartbear:
- Restricting the the source control provider option for the `create-build` command to only accept `github`, `github-enterprise`, `bitbucket`, `bitbucket-server`, `gitlab`, or `gitlab-onpremise`
- If the user enters an invalid provider, they will see an error and the build will not be sent


pkg/utils/source-control-options.go
```diff
@@ -0,0 +1,19 @@
...
+
+func SourceControl(provider string, logger log.Logger) string {
+	if provider == "" {
+		return provider
+	}
+	switch provider {
+	case "github", "github-enterprise", "bitbucket", "bitbucket-server", "gitlab", "gitlab-onpremise":
+		return provider
+	default:
+		logger.Warn(fmt.Sprintf("%s is not an accepted value for the source control provider, proceeding without a provider", provider))
```
joshedney, line 16: @tomlongridge Did we want to proceed without this information or did we want to fail and exit ?

tomlongridge: Fail and exit if they've attempted to provide a value.

pkg/options/options.go
```diff
@@ -39,7 +39,7 @@ type CLI struct {
 		Dsym               upload.Dsym                   `cmd:"" help:"Upload dSYMs for iOS"`
 		UnityAndroid       upload.UnityAndroid           `cmd:"" help:"Upload Android mappings and NDK symbol files from Unity projects"`
 	} `cmd:"" help:"Upload symbol/mapping files"`
-	CreateBuild          CreateBuild          `cmd:"" help:"Provide extra information whenever you build, release, or deploy your application"`
+	CreateBuild          CreateBuild          `cmd:"" help:"Provide extra information whenever you build, release, or deploy your application. If providing source control information, accepted values for provider are: github, github-enterprise, bitbucket, bitbucket-server, gitlab, gitlab-onpremise"`
```
tomlongridge, line 42: We don't need this detail here IMO

```suggestion
	CreateBuild          CreateBuild          `cmd:"" help:"Provide extra information whenever you build, release, or deploy your application"`
```

tomlongridge (commented):
Can you also add a changelog?

tomlongridge (approved):

PR #103, opened 2024-05-20
[PLAT-12131] Add the ability to set the log level 

joshedney:
## Goal

Add the ability to change the `logLevel` via the `--log-level` command line option.
Add validation to the setting of the log level to ensure that only accepted values are passed.

example:

```
$ bugsnag-cli upload react-native-android ... --log-level=fatal
```

Output:
```
[FATAL] failed after 1 attempts. 401 Unauthorized: invalid apiKey
```

## Testing

Covered by CI

### DIFF
--- a/main.go
+++ b/main.go
@@ -32,7 +32,11 @@ func main() {
 			"version": package_version,
 		})
 
-	logger := log.NewLoggerWrapper(commands.Verbose, commands.LogLevel)
+	if commands.Verbose {
+		commands.LogLevel = "debug"
+	}
+
+	logger := log.NewLoggerWrapper(commands.LogLevel)
 
 	// Build connection URI
 	endpoint, err := utils.BuildEndpointUrl(commands.UploadAPIRootUrl, commands.Port)
--- a/main.go
+++ b/main.go
@@ -32,7 +32,7 @@ func main() {
 			"version": package_version,
 		})
 
-	logger := log.NewLoggerWrapper(commands.Verbose)
+	logger := log.NewLoggerWrapper(commands.Verbose, commands.LogLevel)
 
 	// Build connection URI
 	endpoint, err := utils.BuildEndpointUrl(commands.UploadAPIRootUrl, commands.Port)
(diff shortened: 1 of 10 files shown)


pkg/log/logrus-wrapper.go
```diff
@@ -49,14 +49,20 @@ func (f *CustomFormatter) Format(entry *logrus.Entry) ([]byte, error) {
...
 	logger.Formatter = &CustomFormatter{}
 
-	// Set the log level to debug if verbose is true
+	// Set the log level to debug if verbose is true or logLevel is set
 	if verbose {
 		logger.SetLevel(logrus.DebugLevel)
+	} else if logLevel != "" {
+		level, err := logrus.ParseLevel(logLevel)
+		if err != nil {
+			logger.Fatal("Invalid log level")
```
tomlongridge, line 63: I realise it's unlikely to happen as it's validated before, but it'd be worth logging the value of the string.

main.go
```diff
@@ -32,7 +32,7 @@ func main() {
 			"version": package_version,
 		})
 
-	logger := log.NewLoggerWrapper(commands.Verbose)
+	logger := log.NewLoggerWrapper(commands.Verbose, commands.LogLevel)
```
tomlongridge, line 35: The two-parameter solution seems overly complicated - it would probably make more sense to handle the `verbose` parameter here and set the log level to "debug", then it's just `main` that has to worry about it.

pkg/options/options.go
```diff
@@ -14,6 +14,7 @@ type Globals struct {
 	FailOnUploadError bool              `help:"Stops the upload when a mapping file fails to upload to Bugsnag successfully" default:"false"`
 	Version           utils.VersionFlag `name:"version" help:"Print version information and quit"`
 	Verbose           bool              `name:"verbose" help:"Print verbose output"`
+	LogLevel          string            `help:"Set the log level" default:"info"`
```
tomlongridge, line 17: The help text should give the valid values

PR #112, opened 2024-07-16
Stub JS upload command

JamesLindsay0:
## Goal

Add a new no-op command for uploading js sourcemaps.

## Changeset

New command added

## Testing

Manual testing

### DIFF
--- a/pkg/options/options.go
+++ b/pkg/options/options.go
@@ -36,6 +36,7 @@ type CLI struct {
 		DartSymbol         upload.DartSymbolOptions      `cmd:"" help:"Process and upload symbol files for Flutter" name:"dart"`
 		ReactNativeAndroid upload.ReactNativeAndroid     `cmd:"" help:"Upload source maps for React Native Android"`
 		ReactNativeIos     upload.ReactNativeIos         `cmd:"" help:"Upload source maps for React Native iOS"`
+		Js                 upload.JsOptions              `cmd:"" help:"Upload source maps for the web"`
 		Dsym               upload.Dsym                   `cmd:"" help:"Upload dSYMs for iOS"`
 		UnityAndroid       upload.UnityAndroid           `cmd:"" help:"Upload Android mappings and NDK symbol files from Unity projects"`
 	} `cmd:"" help:"Upload symbol/mapping files"`
--- a/pkg/options/options.go
+++ b/pkg/options/options.go
@@ -36,7 +36,7 @@ type CLI struct {
 		DartSymbol         upload.DartSymbolOptions      `cmd:"" help:"Process and upload symbol files for Flutter" name:"dart"`
 		ReactNativeAndroid upload.ReactNativeAndroid     `cmd:"" help:"Upload source maps for React Native Android"`
 		ReactNativeIos     upload.ReactNativeIos         `cmd:"" help:"Upload source maps for React Native iOS"`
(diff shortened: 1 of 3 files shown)


pkg/options/options.go
```diff
@@ -36,6 +36,7 @@ type CLI struct {
 		DartSymbol         upload.DartSymbolOptions      `cmd:"" help:"Process and upload symbol files for Flutter" name:"dart"`
 		ReactNativeAndroid upload.ReactNativeAndroid     `cmd:"" help:"Upload source maps for React Native Android"`
 		ReactNativeIos     upload.ReactNativeIos         `cmd:"" help:"Upload source maps for React Native iOS"`
+		Js                 upload.JsOptions              `cmd:"" help:"Upload source maps for the web"`
```
tomlongridge, line 39: We'll try and make this generic:

```suggestion
		Js                 upload.JsOptions              `cmd:"" help:"Upload JavaScript source maps"`
```

pkg/upload/js.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"fmt"
+
+	"github.com/bugsnag/bugsnag-cli/pkg/log"
+	"github.com/bugsnag/bugsnag-cli/pkg/utils"
+)
+
+type JsOptions struct {
+	VersionName string      `help:"The version of the app that the source map applies to (as set in the JavaScript notifier). Defaults to the version in the package.json file."`
```
tomlongridge, line 11:
```suggestion
	VersionName string      `help:"The version of the app that the source map applies to. Defaults to the version in the package.json file (if found)."`
```

pkg/upload/js.go
```diff
@@ -0,0 +1,40 @@
...
+	BundleUrl   string      `help:"The url of the minified JavaScript file that the source map relates to. Asterisks can be used as a wildcard."`
```
tomlongridge, line 12:
```suggestion
	BundleUrl   string      `help:"URL of the minified JavaScript file that the source map relates to. Asterisks can be used as a wildcard."`
```

pkg/upload/js.go
```diff
@@ -0,0 +1,40 @@
...
+	SourceMap   string      `help:"Path to the source map file. This usually has the .min.js extension." type:"path"`
+	Bundle      string      `help:"Path to the minified JavaScript file that the source map relates to. If this is not provided then the file will be obtained when an error event is received." type:"path"`
+	ProjectRoot string      `help:"The root path of the project on the filesystem where the source map was built. This will be stripped off from the file name in the displayed stacktraces." type:"path"`
```
tomlongridge, line 15:
```suggestion
	ProjectRoot string      `help:"Path of the root directory on the file system where the source map was built. This will be stripped from the file name in the displayed stack traces." type:"path"`
```

pkg/upload/js.go
```diff
@@ -0,0 +1,40 @@
...
+	Path        utils.Paths `arg:"" name:"path" help:"Path to directory or file to upload" type:"path" default:"."`
```
tomlongridge, line 16: Although it makes sense to allow a directory and use file name conventions to find all the bundle/map pairs, I think the use case for specifying a single file isn't clear and will involve a degree of logic to deduce if it's the bundle or map.

Shall we only allow directories here?

```suggestion
	Path        utils.Paths `arg:"" name:"path" help:"Path to a directory of source maps and bundles to upload" type:"path" default:"."`
```

tomlongridge (approved):

PR #113, opened 2024-07-16
Implement upload for a single url

JamesLindsay0:
## Goal

Fully support uploading a single sourcemap with a specified path and URL.

## Testing

Mazerunner tests added for webpack 4 and 5.


pkg/upload/js.go
```diff
@@ -16,6 +24,158 @@ type JsOptions struct {
 	Path        utils.Paths `arg:"" name:"path" help:"Path to a directory of source maps and bundles to upload" type:"path" default:"."`
 }
 
+// Resolve the project root by walking up from the path specified until a package.json is found
+func resolveProjectRoot(jsOptions JsOptions, path string) (string, error) {
```
tomlongridge, line 28: I think we might want to separate out concerns here a bit if the user has explicitly provided a project root, then we should use this string as the projectRoot parameter to the API exactly. I don't think we should conflate this concept with attempts to find the `package.json`; nor resolve the path or navigate for the purposes of the project root value.

In order to find a potential `package.json` I suggest we only look in the current directory, rather than risk finding config files at a higher level (e.g. in a monorepo). I think the principle of least surprise is better here.

pkg/upload/js.go
```diff
@@ -16,6 +24,158 @@ type JsOptions struct {
...
+	// Walk up the folder structure as far as possible
+	for filepath.Dir(checkPath) != checkPath {
+		packageJson := filepath.Join(checkPath, "package.json")
+		if !utils.FileExists(packageJson) {
+			checkPath = filepath.Dir(checkPath)
+			continue
+		}
+		file, err := os.Open(packageJson)
+		if err != nil {
+			return "", fmt.Errorf("unable to open package.json %s", packageJson)
```
tomlongridge, line 66: The approach we've taken for other commands is to _attempt_ to find data from the environment but don't present an error in doing so as parsing the `package.json` is a secondary activity that the user might not be expected – print Info/Warning messages if information is not found in the file, but we shouldn't error out at this point as an error saying "unable to parse package.json" won't make much sense to the user.

So we put the validation for required data at the very last moment, just before the upload, and only then present an error to say "no version found".

pkg/upload/js.go
```diff
@@ -16,6 +24,158 @@ type JsOptions struct {
...
+	err = server.ProcessFileRequest(endpoint+"/sourcemap", uploadOptions, fileFieldData, timeout, retries, sourceMapPath, dryRun, logger)
+
+	if err != nil {
+		logger.Fatal(err.Error())
+	}
+}
+
 func ProcessJs(
 	apiKey string,
 	jsOptions JsOptions,
```
tomlongridge, line 181: Passing a whole command struct isn't something we've done elsewhere – although we've talked about doing it. I'd be happier following our existing pattern of passing the individual options down for now, and we can discuss simplifying the methods by passing a single struct as a future enhancement.

pkg/upload/js.go
JamesLindsay0: If the user has specifically provided a project root then that is used exactly (see the lines below this). The behaviour of walking up to find the package.json is only used as a fallback.

The bugsnag-source-map CLI uses the working directory as the project root if one is not otherwise specified. We could copy that behaviour if you prefer?

pkg/upload/js.go
JamesLindsay0: Sounds reasonable. The API doesn't need a version anyway so we can just skip the version if it is not specified by the user or found.

pkg/upload/js.go
JamesLindsay0: Could you elaborate on the advantage of passing each field in the struct rather than just the struct itself?

pkg/upload/js.go
tomlongridge: Let's stick to the current working directory for both project root default and potential location for the `package.json`.

pkg/upload/js.go
tomlongridge: Consistency with the existing commands in the codebase. I'd rather refactor everything at once in a separate PR than change the style for just this one command now.

pkg/upload/js.go
JamesLindsay0: The bugsnag-source-maps CLI recursively searches upwards to find a package.json in order to find the version. Is this not the desired behaviour here?

tomlongridge: OK - if it's existing behaviour on the old tool, I'm happy to keep it for the package.json search 👍

pkg/upload/js.go
```diff
@@ -16,9 +24,163 @@ type JsOptions struct {
 	Path        utils.Paths `arg:"" name:"path" help:"Path to a directory of source maps and bundles to upload" type:"path" default:"."`
 }
 
+// Resolve the project if it isn't specified using the currend working directory
```
tomlongridge, line 27:
```suggestion
// Resolve the project if it isn't specified using the current working directory
```

pkg/upload/js.go
```diff
@@ -16,9 +24,163 @@ type JsOptions struct {
...
+	}
+	workingDirectory, err := os.Getwd()
+	if err != nil {
+		return path
+	}
+	return workingDirectory
+}
+
+// Attempt to parse information from the package.json file if values aren't provided on the command line
+func ResolveVersion(versionName string, projectRoot string, logger log.Logger) (string, error) {
```
tomlongridge, line 40:
```suggestion
func resolveVersion(versionName string, projectRoot string, logger log.Logger) (string, error) {
```

pkg/upload/js.go
```diff
@@ -16,9 +24,163 @@ type JsOptions struct {
...
+	if versionName != "" {
+		return versionName, nil
+	}
+	checkPath, err := filepath.Abs(projectRoot)
+	if err != nil {
+		return "", fmt.Errorf("unable to make project root an absolute path %s", projectRoot)
```
tomlongridge, line 46: This should be a warning too - if it doesn't exist, we just don't find the package.json.

pkg/upload/js.go
```diff
@@ -16,9 +24,163 @@ type JsOptions struct {
...
+	}
+	// Walk up the folder structure as far as possible
+	for filepath.Dir(checkPath) != checkPath {
+		packageJson := filepath.Join(checkPath, "package.json")
```
tomlongridge, line 50: Does this work without erroring if we don't find a package.json and/or if you don't have permission to ascend above a certain level?

It would be worth adding a test where we don't have a package.json to ensure it doesn't error out.

pkg/upload/js.go
```diff
@@ -16,9 +24,163 @@ type JsOptions struct {
...
+			return "", nil
+		}
+		var parsedPackageJson map[string]interface{}
+		err = json.Unmarshal(byteValue, &parsedPackageJson)
+		if err != nil {
+			logger.Warn(fmt.Sprintf("unable to parse %s: %s", packageJson, err))
+			return "", nil
+		}
+		if parsedPackageJson["version"] == nil {
+			logger.Warn(fmt.Sprintf("missing version in %s", packageJson))
```
tomlongridge, line 72: This feels like more of a Debug message to say "no version found in ..."

pkg/upload/js.go
```diff
@@ -27,12 +189,37 @@ func ProcessJs(
...
+
+		// Dist is the default output path for webpack
+		if utils.IsDir(filepath.Join(path, "dist")) {
+			outputPath = filepath.Join(path, "dist")
+		}
+
+		// Set a default value for projectRoot if it's not defined
+		projectRoot := resolveProjectRoot(projectRoot, path)
+
+		appVersion, err := ResolveVersion(versionName, projectRoot, logger)
```
tomlongridge, line 204: We should keep the concept of "project root" as being _only_ the path passed to the API. For everything else we use the `path`.

The distinction is that the latter is expected to be on the filesystem of the machine running the command; whereas the latter could be separate machine that built the source map and isn't accessible during the upload.

pkg/upload/js.go
```diff
@@ -16,9 +24,163 @@ type JsOptions struct {
...
+		return appVersion, nil
+	}
+	return "", nil
+}
+
+// Attempt to find the source maps by walking the build directory if it is not passed in to the command line
+func ResolveSourceMapPaths(sourceMap string, outputPath string, projectRoot string, logger log.Logger) ([]string, error) {
+	if sourceMap != "" {
+		sourceMap := sourceMap
+		if !path.IsAbs(sourceMap) {
```
tomlongridge, line 86: As discussed IRL, let's not prepend project root and assume that the path provided is relative to the running app.

pkg/upload/js.go
```diff
@@ -27,12 +189,37 @@ func ProcessJs(
...
+		sourceMapPaths, err := ResolveSourceMapPaths(sourceMap, outputPath, projectRoot, logger)
+		if err != nil {
+			return err
+		}
 
-	for _, path := range jsOptions.Path {
-		rootDirPath = path
+		// Check that we now have a source map path
+		if len(sourceMapPaths) == 0 {
+			return fmt.Errorf("could not find a source map, please specify the path by using --source-map or SOURCEMAP_FILE environment variable")
```
tomlongridge, line 217: We don't currently support environment variables in the CLI.

pkg/upload/js.go
```diff
@@ -16,9 +24,163 @@ type JsOptions struct {
...
+	if len(sourceMapPaths) == 0 {
+		logger.Debug(fmt.Sprintf("No main source map found in: %s", outputPath))
+	} else {
+		logger.Debug(fmt.Sprintf("Found source maps: %s", strings.Join(sourceMapPaths, ", ")))
+	}
+	return sourceMapPaths, nil
+}
+
+// Attempt to find the bundle path by changing the extension of the source map if the bundle path is not passed in to the command line
+func ResolveBundlePath(
```
tomlongridge, line 115:
```suggestion
func resolveBundlePath(
```

pkg/upload/js.go
JamesLindsay0: The filepath.join is a string operation and will never fail. If the file is not found or the permissions are not correct then the file.open will fail. This error is caught.

In Go all of the recoverable errors must be passed up as return values and handled.

pkg/upload/js.go
JamesLindsay0: A package.json file must contain a version - https://docs.npmjs.com/creating-a-package-json-file#packagejson-fields.

pkg/upload/js.go
JamesLindsay0: Adding a test for this is difficult because there is a package.json in the root of this project and so if the fixture didn't have one then the root one would be used.

pkg/upload/js.go
```diff
@@ -16,9 +23,160 @@ type JsOptions struct {
...
+}
+
+// Attempt to parse information from the package.json file if values aren't provided on the command line
+func resolveVersion(versionName string, path string, logger log.Logger) string {
+	if versionName != "" {
+		return versionName
+	}
+	checkPath, err := filepath.Abs(path)
+	if err != nil {
+		logger.Warn(fmt.Sprintf("unable to make project root an absolute path %s: %s", path, err))
```
tomlongridge, line 45:
```suggestion
		logger.Warn(fmt.Sprintf("unable to make an absolute path %s: %s", path, err))
```

tomlongridge (approved):
LGTM - just one minor text change

PR #119, opened 2024-07-18
Add multiple JS file upload

JamesLindsay0:
## Goal

Support uploading multiple source map files at the same time.

## Design

The base URL is the same as the old CLI. The path relative to the directory specified is used (matching the old behaviour).

## Testing

A test to check that the URL is properly formatted.


pkg/upload/js.go
```diff
@@ -210,8 +224,21 @@ func ProcessJs(
 			return fmt.Errorf("could not find a source map, please specify the path by using --source-map")
 		}
 
+		// Ensure that exactly one of --bundle-url and --base-url is specified
+		if bundleUrl == "" && baseUrl == "" {
```
tomlongridge, line 228: We should also validate that if the user provides a directory as `path` then the `baseUrl` is given; and conversely if a single file is given then `bundleUrl` is given. Perhaps this supersedes this check?

pkg/upload/js.go
```diff
@@ -150,7 +153,17 @@ func uploadSingleSourceMap(
...
 
-	uploadOptions, err := utils.BuildJsUploadOptions(apiKey, appVersion, bundleUrl, projectRoot, overwrite)
+	url := bundleUrl
+	if baseUrl != "" {
+		relativePath, err := filepath.Rel(outputPath, bundlePath)
+		if err != nil {
+			logger.Warn(fmt.Sprintf("cannot make path relative to create URL, skipping: %v", err))
+			return
+		}
+		url = baseUrl + relativePath
```
tomlongridge, line 163: As discussed, let's make this simpler and just prepend the bundle file name to the base dir - that way it's more obvious to the user how this works and in complicated cases where the structures vary in the build and hosted structure, they can work out what to do.

I'd also suggest that this logic is moved up to the `ProcessJs` function so that at this level it just takes a single file to upload with a URL and doesn't need to know about the `outputPath`.

pkg/upload/js.go
tomlongridge, line 1: Suggest suffixing all parameters that are paths with `Path`, so we can clearly distinguish between paths, URLs and other data. (Internal only, not the option names themselves)

tomlongridge, line 1: Should the appending of `dist` to the root only be done if a path isn't explicitly given? Otherwise it could be confusing to the user.

Also, what if the path is a file?

features/js/js-webpack5.feature
```diff
@@ -65,6 +65,19 @@ Feature: Webpack 5 js Integration Tests
     And the sourcemap payload field "projectRoot" ends with "bugsnag-cli"
     And the sourcemap payload field "overwrite" equals "true"
 
+  Scenario: Base URL correctly appends the path
+    When I run bugsnag-cli with upload js --upload-api-root-url=http://localhost:9339 --api-key=1234567890ABCDEF1234567890ABCDEF --overwrite --base-url=example.com --project-root=features/js/fixtures/js-webpack5 features/js/fixtures/js-webpack5/dist
+    And I wait to receive 1 sourcemaps
```
tomlongridge, line 70: Can we add a multi-file upload for completeness?

pkg/upload/js.go
```diff
@@ -17,6 +17,7 @@ import (
 type JsOptions struct {
 	VersionName string      `help:"The version of the app that the source map applies to. Defaults to the version in the package.json file (if found)."`
 	BundleUrl   string      `help:"URL of the minified JavaScript file that the source map relates to. Asterisks can be used as a wildcard."`
+	BaseUrl     string      `help:"The base URL of the minified files. The relative path is appended onto this for each file. Asterisks can be used as a wildcard."`
```
tomlongridge, line 20:
```suggestion
	BundleUrl   string      `help:"For single file uploads, the URL of the minified JavaScript file that the source map relates to. Asterisks can be used as a wildcard."`
	BaseUrl     string      `help:"For directory-based uploads, the URL of the base directory for the minified JavaScript files that the source maps relate to. The relative path is appended onto this for each file. Asterisks can be used as a wildcard."`
```

pkg/upload/js.go
JamesLindsay0: The appending of dist only happens when the dist path exists. I think it is useful to keep this appending if the user specified the project root instead of the dist folder as the path.

If path is a file then we upload that file (if it is a .map file). I have a test for that case.

tomlongridge: I think if the option is explicitly given then we shouldn't try and amend it - the principle of the CLI is to give customers all the tools to suit their situation and if we try and tack on the extra `/dist` I feel like there's a chance it won't suit some people.

So just do this if no path is given by the user.

Also, if the path is a file I think we'll end up with `myfile.map/dist`? Probably worth guarding.

tomlongridge (approved):

PR #120, opened 2024-07-19
Add sources content

JamesLindsay0:
## Goal

Add the sources content if it is not already populated.

## Changes

I had to modify the uploader to support both on disk files and in memory files.

## Testing

Unit test added.


pkg/upload/js.go
```diff
@@ -108,6 +109,101 @@ func resolveSourceMapPaths(sourceMapPath string, outputPath string, logger log.L
 	return sourceMapPaths, nil
 }
 
+// Reads a JSON source map into memory. Required in order to check if the sourcesContent exists.
+func ReadSourceMap(path string, logger log.Logger) (map[string]interface{}, error) {
+	logger.Debug(fmt.Sprintf("Reading sourcemap %s", path))
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return map[string]interface{}{}, err
```
tomlongridge, line 117: These error cases should probably have a log message associated with them?

pkg/upload/js.go
```diff
@@ -108,6 +109,101 @@ func resolveSourceMapPaths(sourceMapPath string, outputPath string, logger log.L
...
+	}
+	var sourceMapContents map[string]interface{}
+	err = json.Unmarshal(data, &sourceMapContents)
+	if err != nil {
+		return map[string]interface{}{}, err
+	}
+	return sourceMapContents, nil
+}
+
+func addSourcesContent(section map[string]interface{}, relativePath string, logger log.Logger) {
```
tomlongridge, line 127: I previously said we shoud reserve that term for the string we pass to the API only. However it seems justified to call it `projectRoot` here (and in calling function) as this is specifically the project root we want (and not necessarily the base directory of the project, if these are different)

```suggestion
func addSourcesContent(section map[string]interface{}, projectRoot string, logger log.Logger) {
```

pkg/upload/js.go
```diff
@@ -108,6 +109,101 @@ func resolveSourceMapPaths(sourceMapPath string, outputPath string, logger log.L
...
+
+	// Skip if the sources content is already valid
+	untypedSourcesContent, hasSourcesContent := section["sourcesContent"]
+	if hasSourcesContent {
+		sourcesContent, sourcesContentValid := untypedSourcesContent.([]interface{})
+		if sourcesContentValid && len(sourcesContent) == len(sources) {
+			logger.Debug("SourcesContent is already populated")
+			return
+		}
+	}
```
tomlongridge, line 147: Where is this untyped and valid content logic taken from? I couldn't glean it from the old tool or any online spec. Would be good to see a summary of the expected file format, or citation, and/or some more explanatory comments for future readers.

pkg/upload/js.go
```diff
@@ -108,6 +109,101 @@ func resolveSourceMapPaths(sourceMapPath string, outputPath string, logger log.L
...
+			if questionMark-1 > 0 {
+				sourcePath = sourcePath[:questionMark-1]
+			}
+		}
+		if !path.IsAbs(sourcePath) {
+			sourcePath = path.Join(relativePath, sourcePath)
+		}
+		content, err := os.ReadFile(sourcePath)
+		if err != nil {
+			logger.Warn(fmt.Sprintf("Cannot read source file '%s' when searching relative to '%s'.", sourcePath, relativePath))
```
tomlongridge, line 183: `sourcePath` will have been amended to be absolute by here if it's relative, so I don't think the error message will read logically. How about just:


```suggestion
			logger.Warn(fmt.Sprintf("Cannot read referenced source file '%s'", sourcePath))
```

pkg/upload/js.go
```diff
@@ -146,7 +242,22 @@ func uploadSingleSourceMap(
...
+
+	AddSources(sourceMapContents, projectRoot, logger)
+	if err != nil {
+		logger.Fatal(err.Error())
+	}
+
+	encodedSourceMap, err := json.Marshal(sourceMapContents)
+	if err != nil {
+		logger.Fatal(err.Error())
+	}
```
tomlongridge, line 258: I don't know what the likleihood is of an error unmarshalling the JSON is, but I think I'd feel happier if we used the original content if we detect the `sourcesContent` is already present; and also if there was an error in the un/marshalling itself. So if it was an unexpected file format, we just continue to upload with a warning that no sources were attached.

I'm open to other views on this though. @imjoehaines/@lemnik - any thoughts here on the most desirable behaviour?

pkg/upload/js.go
JamesLindsay0: This logic is based on the JS implementation (but that obviously didn't have types) as well as the sourcemap specification https://bit.ly/sourcemap which dictates the types. I have added in a comment at the top of that function

tomlongridge (approved):

PR #121, opened 2024-07-24
JS upload command

JamesLindsay0:
## Goal & design

See https://smartbear.atlassian.net/wiki/spaces/EN/pages/4299292809/PLAT-12409+ED+-+Add+browser+source+map+uploads+to+CLI

## Testing

New maze-runner and unit tests.


pkg/upload/js.go
```diff
@@ -0,0 +1,376 @@
...
+		logger.Fatal(err.Error())
+	}
+
+	var sourceMapFile server.FileField
+
+	sourceMapModified := AddSources(sourceMapContents, sourceMapPath, projectRoot, logger)
+	if sourceMapModified {
+		encodedSourceMap, err := json.Marshal(sourceMapContents)
+		if err != nil {
+			logger.Fatal(err.Error())
```
tomlongridge, line 271: I think more context here would be useful:

```suggestion
			logger.Fatal(fmt.Sprintf("Failed generate valid source map JSON with original sources added: %s", err.Error()))
```

pkg/upload/js.go
```diff
@@ -0,0 +1,376 @@
...
+			return fmt.Errorf("`--bundle-url` must not be set when uploading from a directory")
+		}
+
+		// Add a slash if it is not already on the end of the base URL
+		if len(baseUrl) > 0 && baseUrl[len(baseUrl)-1] != '/' {
+			baseUrl += "/"
+		}
+
+		for _, sourceMapPath := range sourceMapPaths {
+			uploadSingleSourceMap(bundleUrl, baseUrl, bundlePath, sourceMapPath, apiKey, appVersion, projectRoot, endpoint, timeout, retries, overwrite, dryRun, logger)
```
tomlongridge, line 370: For consistency with other commands, we should pass any error up for potential handling higher up. The idea of the CLI is to provide commands as building blocks, so we want to allow an umbrella command to call this and potentially handle an error.

Where there are potentially multiple files (e.g. dSYMs), we should exit out of the loop if there's an error. By calling fatal error you're doing this now, but we'll want to check the error returned and return it up the callstack.

features/js/fixtures/js-multiple-maps/dist/main.js
gingerbenw, line 1: Should the `dist` directory be in `.gitignore` ? I wouldn't usually check in any generated files.

joshedney: For this fixture, its probably fine. As this is a fixed "where we expect things to be" fixture and the `base-fixtures` is where we generate it from scratch.

joshedney:
Don't forget to add an entry to the `CHANGELOG.md`

tomlongridge (approved):

PR #125, opened 2024-07-29
[PLAT-12381] BugSnag CLI installation CI tests

joshedney:
## Goal

Test the installation of the BugSnag CLI via the various package managers we support.

## Changeset

Add MR tests for installation via the supported package managers to ensure they are covered.

## Testing

Covered by CI


features/steps/steps.rb
```diff
@@ -172,3 +170,42 @@ def get_version_number(file_path)
...
+  @output = `npm install #{@bugsnag_cli_package_path}`
+end
+
+Then(/^I install the bugsnag\-cli via yarn in the new project$/) do
+  @output = `yarn add #{@bugsnag_cli_package_path}`
+end
+
+Then(/^I install the bugsnag\-cli via pnpm in the new project$/) do
+  @output = `pnpm add #{@bugsnag_cli_package_path}`
+end
```
tomlongridge, line 198: Shouldn't the `init` be `yarn init` for the second step?

How about combining these either into 1 step or 3 separate ones that create the directory, init it and add the CLI?

`I install the bugsnag\-cli using {string} in a new directory`

features/steps/steps.rb
```diff
@@ -172,3 +170,42 @@ def get_version_number(file_path)
...
+
+Then(/^I clean up the project$/) do
+  # Change directory to feature/cli
+  Dir.chdir(@base_dir)
+  # Remove the fixture directory
+  FileUtils.rm_rf("#{Dir.pwd}/features/cli/fixture")
+  # Remove the bugsnag-cli package
+  FileUtils.rm(@bugsnag_cli_package_path)
+end
```
tomlongridge, line 211: Would this be necessary if you use unique names for the fixtures?

features/steps/steps.rb
```diff
@@ -172,3 +170,42 @@ def get_version_number(file_path)
...
+
+Then(/^I install the bugsnag\-cli via yarn in the new project$/) do
+  @output = `yarn add #{@bugsnag_cli_package_path}`
+end
+
+Then(/^I install the bugsnag\-cli via pnpm in the new project$/) do
+  @output = `pnpm add #{@bugsnag_cli_package_path}`
+end
+
+Then('the node_modules bin directory should contain {string}') do |arg|
```
tomlongridge, line 200: Suggest generalising this step a little too - the test would be equally readable:

```suggestion
Then('the {string} directory should contain {string}') do |arg|
```

features/steps/steps.rb
```diff
@@ -172,3 +170,39 @@ def get_version_number(file_path)
...
+  @output = `npm pack`
+  @bugsnag_cli_package_path = "#{@base_dir}/#{@output}"
+end
+
+When('I install the bugsnag-cli via {string} in a new directory') do |package_manager|
+  case package_manager
+  when 'npm'
+    @fixture_dir = "#{@base_dir}/features/cli/fixtures/#{package_manager}"
+    Dir.mkdir(@fixture_dir)
+    Dir.chdir(@fixture_dir)
```
tomlongridge, line 185: This can be moved outside the `case` block I think?

(And the final `chdir`)

features/steps/steps.rb
```diff
@@ -172,3 +170,39 @@ def get_version_number(file_path)
...
+    @init_output = `pnpm init -y`
+    @install_output = `pnpm add #{@bugsnag_cli_package_path}`
+    Dir.chdir(@base_dir)
+  end
+end
+
+Then('the {string} directory should contain {string}') do |directory, package|
+  Maze.check.include(`ls #{@fixture_dir}/#{directory}`, package)
```
tomlongridge, line 207: You're not actually using the directory name

features/steps/steps.rb
joshedney: I've refactored this block of code to try and make it as DRY as possible.

features/steps/steps.rb
joshedney: As far as I can tell I'm using the directory name. 

`Maze.check.include(`ls #{@fixture_dir}/#{directory}`, package)` translates to `Maze.check.include(`ls features/cli/fixtures/npm/node_modules/.bin/`, bugsnag-cli)`

tomlongridge: Sorry - not sure what I was thinking. I'm sure it still had `node_modules` in it when I wrote that comment 🤦

tomlongridge (approved):

PR #157, opened 2024-12-03
[PLAT -13263] Create the `xcode-archive` command 

joshedney:
## Goal

Add the `xcode-archive` command to handle DSYM's produced by an xcode archive.

- The command accepts the following
- - .xcarchive file
- - a path containing xcarchives
- - an Xcode project ( it will workout the path to the xcode archive location)

## Testing

Tested locally - tests to be added


pkg/upload/xcode-archive.go
```diff
@@ -0,0 +1,173 @@
...
+	if err != nil {
+		return err
+	}
+	if len(dwarfInfo) == 0 {
+		return fmt.Errorf("No dSYM files found in: %s", xcarchivePath)
+	}
+	logger.Info(fmt.Sprintf("Found %d dSYM files in %s", len(dwarfInfo), xcarchivePath))
+
+	// Locate Info.plist if not explicitly provided
+	if plistPath == "" && options.ApiKey == "" && buildSettings != nil {
```
tomlongridge, line 123: IMO we shouldn't use the project settings (buildSettings) for anything other than locating the archive file as there's a reasonable expectation that we're doing the upload completely from the archive contents.

Am I right in thinking that `Info.plist` file is at the top of an archive? In which case, we should read this unless the plist path is specified in the command.

pkg/upload/xcode-archive.go
```diff
@@ -0,0 +1,173 @@
...
+		xcarchivePath,
+		xcarchiveOptions.Shared.IgnoreEmptyDsym,
+		xcarchiveOptions.Shared.IgnoreMissingDwarf,
+		logger,
+	)
+	tempDirs = append(tempDirs, tempDir)
+	if err != nil {
+		return err
+	}
+	if len(dwarfInfo) == 0 {
```
tomlongridge, line 117: Can we abstract the rest of this into a shared function with xcode-build to ensure they don't diverge?

I imagine it could take the plist path and the dsym directory and perform the uploads?

pkg/utils/files.go
```diff
@@ -53,110 +75,170 @@ func BuildFileList(paths []string) ([]string, error) {
...
-	tempDir, err := os.MkdirTemp("", fmt.Sprintf("bugsnag-cli-%s-unpacking-*", slug))
+// FindLatestFolder searches for the most recently modified folder in a given directory.
+//
+// Parameters:
+// - directory (string): The directory to search in.
+//
+// Returns:
+// - string: The path to the newest folder.
+// - error: Any error encountered during the search.
+func FindLatestFolder(directory string) (string, error) {
```
tomlongridge, line 162: Did you mean to use this above?

pkg/ios/xcarchive.go
```diff
@@ -0,0 +1,128 @@
...
+	if err != nil {
+		// If the command fails, check stderr for additional details
+		if strings.Contains(stderr.String(), "does not exist") {
+			// If the key is not set, return the default location for Xcode archives
+			usr, err := user.Current()
+			if err != nil {
+				return "", fmt.Errorf("unable to get current user: %w", err)
+			}
+			// Return default archive location under user's home directory
+			return filepath.Join(usr.HomeDir, "Library", "Developer", "Xcode", "Archives"), nil
```
tomlongridge, line 44: Should we do an exists check here, just to make sure?

pkg/ios/xcarchive.go
```diff
@@ -0,0 +1,128 @@
...
+	path, err := findLatestXCArchive(path, scheme)
+	if err != nil {
+		return "", err
+	} else if path == "" {
+		// If no archive is found, return an error indicating the folder doesn't contain any matching archives
+		return "", fmt.Errorf("No xcarchive found in %s", path)
+	}
+	// Return the path of the latest archive found
+	return path, nil
+}
```
tomlongridge, line 82: Is this function providing any purpose, other than wrapping `findLatestXCArchive`? Feels like they could easily be combined.

pkg/upload/xcode-archive.go
```diff
@@ -0,0 +1,102 @@
...
+						return fmt.Errorf("Error determining default scheme: %w", err)
+					}
+				}
+
+				archiveLocation, err := ios.GetXcodeArchiveLocation()
+				if err != nil {
+					logger.Warn(fmt.Sprintf("Error getting Xcode archive location: %s", err))
+					return err
+				}
+				xcarchivePath, err = ios.GetLatestXcodeArchive(archiveLocation, xcarchiveOptions.Scheme)
```
tomlongridge, line 63: I can't see you ever needing to call one of these without the other - how about just combining them?

pkg/upload/xcode-archive.go
```diff
@@ -0,0 +1,102 @@
...
+				if xcarchiveOptions.Scheme == "" {
+					xcarchiveOptions.Scheme, err = ios.GetDefaultScheme(path)
+					if err != nil {
+						return fmt.Errorf("Error determining default scheme: %w", err)
+					}
+				}
+
+				archiveLocation, err := ios.GetXcodeArchiveLocation()
+				if err != nil {
+					logger.Warn(fmt.Sprintf("Error getting Xcode archive location: %s", err))
```
tomlongridge, line 60: Can you consolidate how you refer to archives? There's a mixture of .xcarchive, xcarchive and Xcode archive in messages.

Suggest "xcarchive"?

pkg/upload/xcode-archive.go
```diff
@@ -0,0 +1,102 @@
...
+					return err
+				}
+				xcarchivePath, err = ios.GetLatestXcodeArchive(archiveLocation, xcarchiveOptions.Scheme)
+				if err != nil {
+					return fmt.Errorf("Error locating latest .xcarchive: %w", err)
+				}
+			} else {
+				return fmt.Errorf("No .xcarchive found in %s", path)
+			}
+		}
```
tomlongridge, line 70: Would probably be suitable to have a log info here to confirm the archive file we've picked

pkg/options/options.go
```diff
@@ -73,6 +73,17 @@ type XcodeBuild struct {
...
+type XcodeArchive struct {
+	Path               utils.Paths `arg:"" name:"path" help:"The path to the directory or file to upload" type:"path" default:"."`
+	IgnoreEmptyDsym    bool        `help:"Throw warnings instead of errors when a dSYM file is found, rather than the expected dSYM directory"`
+	IgnoreMissingDwarf bool        `help:"Throw warnings instead of errors when a dSYM with missing DWARF data is found"`
+	Scheme             string      `help:"The name of the Xcode options.Scheme used to build the application"`
+	VersionName        string      `help:"The version of the application"`
+	XcodeProject       utils.Path  `help:"The path to an Xcode project, workspace or containing directory from which to obtain build information" type:"path"`
+	ProjectRoot        string      `help:"The path to strip from the beginning of source file names referenced in stacktraces on the BugSnag dashboard" type:"path"`
+	Configuration      string      `help:"The configuration used to build the application"`
+}
```
tomlongridge, line 85: This looks identical to xcode-build, I don't think we use them all do we? e.g. XcodeProject

It would be good if they could inherit a common superset though perhaps?

The Path text could do with more detail on what you can pass in

pkg/upload/xcode-build.go
```diff
@@ -2,159 +2,134 @@ package upload
...
+		// Locate Info.plist if not already specified
+		if plistPath == "" && options.ApiKey == "" && buildSettings != nil {
+			plistPathExpected := filepath.Join(buildSettings.ConfigurationBuildDir, buildSettings.InfoPlistPath)
+			if utils.FileExists(plistPathExpected) {
+				plistPath = plistPathExpected
+				logger.Debug(fmt.Sprintf("Found Info.plist at: %s", plistPath))
+			} else {
+				logger.Debug(fmt.Sprintf("Info.plist not found at: %s", plistPathExpected))
 			}
 		}
```
tomlongridge, line 127: Can `ProcessDsymUpload` do the plist path exists check, so you just need to pass in where you think it is?

pkg/ios/dsym-utils.go
```diff
@@ -99,47 +113,52 @@ func FindDsymsInPath(path string, ignoreEmptyDsym, ignoreMissingDwarf bool, logg
...
+// Parameters:
+// - path: The directory path containing the DWARF file.
+// - fileName: The name of the DWARF file to be analyzed.
+//
+// Returns:
+// - A slice of DwarfInfo structs containing extracted DWARF information.
 func getDwarfFileInfo(path, fileName string) []*DwarfInfo {
 	var dwarfInfo []*DwarfInfo
-
-	cmd := exec.Command(utils.DWARFDUMP, "-u", strings.TrimSuffix(fileName, ".zip"))
```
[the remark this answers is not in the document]
joshedney: Correct, as we're not dealing with the zip file but the extracted files/folder it won't have the zip file ext.

pkg/ios/xcarchive.go
joshedney: It wouldn't hurt

pkg/upload/xcode-build.go
joshedney: It looks like ProcessDsymUpload is already doing a `FileExists` check so this is not needed.

pkg/upload/xcode-archive.go
joshedney: I've adjusted the looping through the provided paths to match other functions which now means the log for showing what file we're using is included.

pkg/upload/xcode-archive.go
joshedney: I think I've gone through and updated the relevant places so that they all should now be `xcarchive`

pkg/upload/xcode-archive.go
joshedney: These have been consolidated

pkg/ios/xcarchive.go
joshedney: You are correct, I've combined the two

pkg/ios/plist-reader-json.go
```diff
@@ -20,31 +22,60 @@ type bugsnagProjectDetails struct {
...
+	var plistData PlistData
+	err = json.Unmarshal(output, &plistData)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to unmarshal plist data into PlistData struct")
+	}
+
+	// Validate required fields in the plist data
+	if plistData.VersionName == "" || plistData.BundleVersion == "" {
+		return nil, fmt.Errorf("invalid plist data: missing required fields (VersionName or BundleVersion)")
+	}
```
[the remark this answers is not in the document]
joshedney: Having the validation there means we do not have to validate across multiple files. In other areas we've tried to be as DRY as possible.

pkg/ios/xcarchive.go
```diff
@@ -0,0 +1,99 @@
...
+			if strings.Contains(stderr.String(), "does not exist") {
+				// If the key is not set, return the default location for xcarchives
+				usr, err := user.Current()
+				if err != nil {
+					return "", fmt.Errorf("unable to get current user: %w", err)
+				}
+				// Return default archive location under user's home directory
+				archivePath := filepath.Join(usr.HomeDir, "Library", "Developer", "Xcode", "Archives")
+				if utils.IsDir(archivePath) {
+					return filepath.Join(usr.HomeDir, "Library", "Developer", "Xcode", "Archives"), nil
```
tomlongridge, line 48:
```suggestion
					return archivePath, nil
```

tomlongridge (approved):
LGTM - just one tweak

PR #172, opened 2025-02-10
[PLAT-13629] Get NDK version from the properties file

joshedney:
## Goal

Ensure that we get the Android NDK version from a reliable source.

## Changeset

Get the `Pkg.Revision` from the NDK `source.properties` file

## Testing

Covered by CI

### DIFF
--- a/pkg/android/get-ndk-version.go
+++ b/pkg/android/get-ndk-version.go
@@ -1,21 +1,52 @@
 package android
 
 import (
+	"bufio"
+	"fmt"
+	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
 )
 
 // GetNdkVersion - Returns the major NDK version
-func GetNdkVersion(path string) (int, error) {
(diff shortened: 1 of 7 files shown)


pkg/android/get-ndk-version.go
```diff
@@ -1,21 +1,52 @@
...
+	defer file.Close()
 
-	ndkIntVersion, err := strconv.Atoi(ndkVersion[0])
+	// Read the file line by line
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		line := scanner.Text()
+		if strings.HasPrefix(line, "Pkg.Revision") {
+			// Extract version string
+			parts := strings.Split(line, " = ")
```
lemnik, line 30: Technically the format doesn't require the spaces - it would be better to split on just the "=" and then trim the resulting string.

```suggestion
			parts := strings.Split(line, "=")
```

PR #173, opened 2025-02-11
[PLAT-13621] React Native iOS Rework

joshedney:
## Goal

Rework the `upload react-native-ios` command to work with Xcode Archives rather than builds by default.

Example:
```
bugsnag-cli upload react-native-ios /path/to/react-native-project
```

Output:
```
[INFO] Uploading main.jsbundle.map to http://localhost:9339/react-native-source-map
[INFO] Uploaded main.jsbundle.map
```

Pointing the command to a standalone `.xcarchive` also works, but you will need to provide additional options using the command escape hatches.

## Testing

Tested using the newly generated fixtures.

Escape hatches have been tested locally using the generated fixture

CI tests for react native have been slimmed down for now to cover the happy path options. Additional CI tests will be added to cover all command escape hatches in a separate PR

Covered by CI


pkg/upload/react-native-ios.go
```diff
@@ -2,45 +2,46 @@ package upload
...
 func ProcessReactNativeIos(options options.CLI, endpoint string, logger log.Logger) error {
 	iosOptions := options.Upload.ReactNativeIos
-	var rootDirPath string
-	var plistData *ios.PlistData
-	var buildSettings *ios.XcodeBuildSettings
-	var err error
+	var (
+		rootDirPath string
+		plistData   *ios.PlistData
+		//buildSettings    *ios.XcodeBuildSettings
```
tomlongridge, line 20: Remove

pkg/upload/react-native-ios.go
```diff
@@ -68,123 +71,118 @@ func ProcessReactNativeIos(options options.CLI, endpoint string, logger log.Logg
...
-					buildSettings, err = ios.GetXcodeBuildSettings(iosOptions.Ios.XcodeProject, iosOptions.Ios.Scheme, "")
-					if err != nil {
-						logger.Warn(err.Error())
-					}
-				}
 			} else {
 				return fmt.Errorf("Could not find an Xcode project file, please specify the path by using --xcode-proj-path")
 			}
 
+			logger.Debug(fmt.Sprintf("Found Xcode Scheme: %s", iosOptions.Ios.Scheme))
```
tomlongridge, line 78:
```suggestion
			logger.Debug(fmt.Sprintf("Found Xcode scheme: %s", iosOptions.Ios.Scheme))
```

pkg/upload/react-native-ios.go
```diff
@@ -68,123 +71,118 @@ func ProcessReactNativeIos(options options.CLI, endpoint string, logger log.Logg
...
+
+			xcodeArchivePath, err = ios.GetLatestXcodeArchiveForScheme(iosOptions.Ios.Scheme)
+
+			if err != nil {
+				return fmt.Errorf("Error locating latest xcarchive: %w", err)
+			}
+
+		} else {
+			return fmt.Errorf("No xcarchive found in specified paths")
```
tomlongridge, line 87: Would it be more appropriate to say "path should be an .xcarchive or the directory of your Xcode project"? Probably useful to log the path for clarity too.

pkg/upload/react-native-ios.go
```diff
@@ -68,123 +71,118 @@ func ProcessReactNativeIos(options options.CLI, endpoint string, logger log.Logg
...
 			} else {
 				return fmt.Errorf("Could not find an Xcode project file, please specify the path by using --xcode-proj-path")
 			}
 
+			logger.Debug(fmt.Sprintf("Found Xcode Scheme: %s", iosOptions.Ios.Scheme))
+
+			xcodeArchivePath, err = ios.GetLatestXcodeArchiveForScheme(iosOptions.Ios.Scheme)
+
+			if err != nil {
+				return fmt.Errorf("Error locating latest xcarchive: %w", err)
```
tomlongridge, line 83: Shall we say "...from Xcode project (scheme XXX), please specify the archive path directly using --archive-path"

pkg/upload/react-native-ios.go
```diff
@@ -68,123 +71,118 @@ func ProcessReactNativeIos(options options.CLI, endpoint string, logger log.Logg
...
+		return fmt.Errorf("Could not find a bundle file, please specify the path by using --bundle-path")
+	}
 
-		// Check that we now have a bundle path
-		if iosOptions.ReactNative.Bundle == "" {
-			return fmt.Errorf("Could not find a bundle file, please specify the path by using --bundle-path")
+	// Check that the source map file exists and error out if it doesn't
+	if iosOptions.ReactNative.SourceMap != "" {
+		if !utils.FileExists(iosOptions.ReactNative.SourceMap) {
+			return fmt.Errorf("unable to find specified source map: %s", iosOptions.ReactNative.SourceMap)
```
tomlongridge, line 138:
```suggestion
			return fmt.Errorf("Unable to find specified source map: %s", iosOptions.ReactNative.SourceMap)
```

pkg/upload/react-native-ios.go
```diff
@@ -2,45 +2,46 @@ package upload
...
-		rootDirPath = path
-		if !utils.FileExists(buildDirPath) {
-			buildDirPath = filepath.Join(path, "build")
-
-			if utils.FileExists(buildDirPath) {
-				rootDirPath = filepath.Join(path, "..")
-
-			} else if iosOptions.ReactNative.Bundle == "" || iosOptions.ReactNative.SourceMap == "" {
-				return fmt.Errorf("unable to find bundle files or source maps in within %s", path)
+		if filepath.Ext(path) == ".xcarchive" {
```
tomlongridge, line 27: I would really like to see this code extracted and shared with the dsym commands – something like "findLatestArchive" that takes the path and returns the archive path using this procedure. (I wonder if there's also an argument to say that we should search a directory for an archive file first, before looking for a project file.)

Then we can be sure that both bundles and dSYMs will follow the same logic.

pkg/options/options.go
```diff
@@ -124,10 +124,11 @@ type ReactNativeShared struct {
...
 type ReactNativeIosSpecific struct {
-	BundleVersion string `help:"The bundle version of this build of the application (Apple platforms only)"`
-	Plist         string `help:"The path to a .plist file from which to obtain build information" type:"path"`
-	Scheme        string `help:"The name of the Xcode options.Ios.Scheme used to build the application"`
-	XcodeProject  string `help:"The path to an Xcode project, workspace or containing directory from which to obtain build information" type:"path"`
+	BundleVersion string     `help:"The bundle version of this build of the application (Apple platforms only)"`
+	Plist         string     `help:"The path to a .plist file from which to obtain build information" type:"path"`
+	Scheme        string     `help:"The name of the Xcode options.Ios.Scheme used to build the application"`
+	XcodeProject  string     `help:"The path to an Xcode project, workspace or containing directory from which to obtain build information" type:"path"`
+	ArchivePath   utils.Path `help:"The path to the Xcarchive file to process" type:"path"`
```
tomlongridge, line 131: Would `--xcarchive-path` be more precise here?

makefile
```diff
@@ -110,14 +110,14 @@ features/base-fixtures/rn0_69/ios:
...
-	cd $@ && xcodebuild -workspace rn0_69.xcworkspace -scheme rn0_69 -configuration Release -sdk iphoneos build
+	cd $@ && xcodebuild -workspace rn0_69.xcworkspace -scheme rn0_69 -configuration Release -allowProvisioningUpdates archive
 
-.PHONY: features/base-fixtures/rn0_69/ios/archive
-features/base-fixtures/rn0_69/ios/archive:
-	cd features/base-fixtures/rn0_69/ && npm i
-	cd features/base-fixtures/rn0_69/ios/ && bundle install
-	cd features/base-fixtures/rn0_69/ios/ && bundle exec pod install --repo-update
-	cd features/base-fixtures/rn0_69/ios/ && xcrun xcodebuild -scheme rn0_69 -workspace rn0_69.xcworkspace -configuration Release -archivePath "../rn0_69.xcarchive" -allowProvisioningUpdates archive
+#.PHONY: features/base-fixtures/rn0_69/ios/archive
```
tomlongridge, line 115: Delete if we're removing these tests?

joshedney:
> > Additional CI tests will be added to cover all command escape hatches.
> 
> Is this coming in a subsequent PR or after this first review?

The plan is for this to come in a subsequent PR

pkg/upload/react-native-ios.go
tomlongridge: Discussed in IRL - it's probably not a good idea to search a directory for an archive as well as it just complicates the possibilities. And that means the code is as abstracted as it needs to be.

tomlongridge (approved):

PR #185, opened 2025-03-24
Update react-native-ios.go

Gregoirevda:
## Goal
When a custom scheme is passed, it will currently pickup the "Debug" build settings. This passes the configuration based on the `dev` flag


mclack:
Hi @Gregoirevda 

Thanks for raising this PR.

This seems like a sensible change, so we'll make sure to put this through our full review process when priorities allow. We'll post any updates here as they become available.

PR #188, opened 2025-03-27
[HOTFIX] Find source path relative to the sourcemap path

joshedney:
## Goal

Remove the `Path` package and replace it with the `filepath` package for consistency throughout the repo.

Currently, If the `sourcePath` is an absolute path we attempt to join it with the  `projectRoot`. This would sometimes result in an incorrect path and we'd be unable to read the source file.

Instead, we should find the `sourcePath` relative to the `sourceMapPath` if the `sourcePath` is an absolute path.

## Testing

Covered by CI

### DIFF
--- a/pkg/upload/js.go
+++ b/pkg/upload/js.go
@@ -170,7 +170,7 @@ func ReadSourceMap(path string, logger log.Logger) (map[string]interface{}, erro
 }
 
 // Based on the source map specification https://bit.ly/sourcemap. Returns if the source map was modified.
-func addSourcesContent(section map[string]interface{}, sourceMapPath string, projectRoot string, logger log.Logger) bool {
+func addSourcesContent(section map[string]interface{}, sourceMapPath string, logger log.Logger) bool {
 	untypedSources, hasSources := section["sources"]
 	if !hasSources {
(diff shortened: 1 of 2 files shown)


pkg/upload/js.go
```diff
@@ -222,8 +221,9 @@ func addSourcesContent(section map[string]interface{}, sourceMapPath string, pro
 				sourcePath = sourcePath[:questionMark-1]
 			}
 		}
-		if !path.IsAbs(sourcePath) {
-			sourcePath = path.Join(projectRoot, sourcePath)
```
tomlongridge, line 226: I think you can remove `projectRoot` from param list and further up the call chain?

pkg/upload/js.go
```diff
@@ -247,13 +247,13 @@ func AddSources(sourceMapContents map[string]interface{}, sourceMapPath string,
 		if sources, ok := sections.([]map[string]interface{}); ok {
```
tomlongridge, line 247: Project root can be removed from `AddSources` too

joshedney: Ahh yes, this has been removed now.

tomlongridge (approved):

PR #192, opened 2025-04-04
README update & Update the react native command to use the dsym command

joshedney:
## Goal

Update the README to match the changes in the proposed docs. 

Update the `upload react-native` command to use the `upload dsym` command instead of the `upload xcode-build` command.

## Testing

Covered by CI

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -4,7 +4,7 @@
 
 ### Breaking changes
 
-We now prioritize Xcode archives over Xcode builds when uploading dSYMs. This means that if you have both an archive and a build, the CLI will upload the dSYMs from the archive when using the `upload dsym` and `upload react-native` commands.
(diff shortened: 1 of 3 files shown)


UPGRADING.md
```diff
@@ -2,6 +2,12 @@
 
 ## v2.x to v3.x
 
+### Breaking changes
+
+We now prioritize Xcode archives over Xcode builds when uploading dSYMs. This means that if you have both an archive and a build, the CLI will search the default archive location set in Xcode for the latest archive for that day and upload the dSYMs from the archive when using the `upload dsym` and `upload react-native` commands.
```
tomlongridge, line 7:
```suggestion
When using the `upload dsym` and `upload react-native` commands, Xcode archives are now prioritized over dSYMs found inside the project from a build. This means that if you have both an archive and a build, the CLI will search the default archive location set in Xcode for the latest archive for that day and upload the dSYMs from the archive.

The `upload xcode-build` commands can be used if you want to always pick dSYM files from a project build instead.
```

tomlongridge (approved):

PR #205, opened 2025-05-13
Update the CLI wrapper to handle spaces in options

joshedney:
## Goal

Update the CLI wrapper to ensure that options with spaces in them are correctly passed to the CLI 

## Testing

Coverd by CI


gingerbenw (approved):
a unit test would be good here, but looks like it resolves the issue

PR #210, opened 2025-06-12
Remove `--overwrite` from the upload dsym commands

joshedney:
## Goal

Remove the `--overwrite` option from the dSYM related commands as this option is not used by the API

## Testing

Covered by CI


tomlongridge (approved):
Yep, it's annoying from a code point of view but I think the right thing to do.

PR #212, opened 2025-06-23
Add support for Unity Android line mapping files

joshedney:
## Goal

- Add support for Unity Android line mapping files
- Update tests to cover new functionality

- Refactor the Xcode build function to allow the Unity iOS command to use parts of it to find the dsyms.
- Add functionality to find the line mappings for an iOS build.
- Add the `unity-ios` command for finding dsyms and line mapping files as part of an Xcode build phase and standalone.
- Add tests to cover the new Unity iOS command.

## Testing

Covered by CI


pkg/options/unity.go
```diff
@@ -0,0 +1,24 @@
+package options
+
+import "github.com/bugsnag/bugsnag-cli/pkg/utils"
+
+// UnityLineMapping is used to specify options for uploading a Unity il2cpp mapping file.
+type UnityLineMapping struct {
+	NoUploadIl2cppMappingFile bool       `help:"Do not upload the il2cpp mapping file"`
+	UploadIl2cppMappingFile   utils.Path `help:"The path to the il2cpp mapping file to upload"`
```
tomlongridge, line 8: Is there anyway to stop the option being called `--no-upload-il-2-cpp-mapping-file` and put the `il2cpp` together?

If not, I'd rather call this `NoUploadLineNumberMappingsFile` / `LineNumberMappingsFile`

pkg/unity/find-line-mappings.go
```diff
@@ -0,0 +1,50 @@
...
+//
+//	path        - an optional explicit path to the mapping file. If provided, it is returned directly.
+//	projectRoot - the root path of the Unity project, used to search for backup artifacts.
+//
+// Returns:
+//
+//	mappingPath - the resolved path to LineNumberMappings.json, or an empty string if not found.
+//	error       - non-nil if there was an error during backup folder resolution.
+func GetAndroidLineMapping(path string, buildDir string) (string, error) {
+	if path != "" {
```
tomlongridge, line 29: I'm not sure I see the point of this – but perhaps we should allow this option to be a directory in which we look for `LineNumberMappings.json`?

pkg/unity/find-line-mappings.go
```diff
@@ -0,0 +1,50 @@
...
+		return path, nil
+	}
+
+	// Check default artifacts path
+	defaultPath := filepath.Join(buildDir, "Library", "Bee", "artifacts", "Android", "il2cppOutput", "cpp", "Symbols", "LineNumberMappings.json")
```
tomlongridge, line 34: Do we understand what "Bee" is in this structure? I just wonder how likely it is to change. Perhaps we should recursive-search the "Library" and "BackUpThisFolder_ButDontShipItWithYourGame" subdirectories and return the first of that name?

pkg/unity/process-line-mappings.go
```diff
@@ -0,0 +1,45 @@
...
+) error {
+	opts := utils.UnityLineMappingOptions{
+		APIKey:         manifestData["apiKey"],
+		AppID:          manifestData["applicationId"],
+		AppVersion:     manifestData["versionName"],
+		AppVersionCode: manifestData["versionCode"],
+		SOBuildID:      soBuildId,
+		ProjectRoot:    options.Upload.UnityAndroid.ProjectRoot,
+		Overwrite:      options.Upload.UnityAndroid.Overwrite,
+	}
```
tomlongridge, line 26: This pattern of having a new struct seems a little convoluted as you have to do the conversion here, just for the build function to take the values out and put them into a map immediately afterwards.

I suggest keeping the existing pattern, at the expense of a relatively long parameter list.

pkg/upload/unity-android.go
```diff
@@ -108,6 +122,28 @@ func ProcessUnityAndroid(globalOptions options.CLI, endpoint string, logger log.
 				if filepath.Base(file) == "libil2cpp.sym.so" && utils.ContainsString(fileList, "libil2cpp.dbg.so") {
 					continue
 				}
+
+				// If we're uploading a libil2cpp.so file, we want to extract the build ID so that we can upload the linemapping file
+				if filepath.Base(file) == "libil2cpp.so" && !unityOptions.Shared.NoUploadIl2cppMappingFile {
```
tomlongridge, line 127: The ED says we'll upload this after the .so files, but that if we can't get a build ID then we'll error out.

Suggest splitting this logic to get the path and build ID here (or error out), then upload the .so files, then upload the mapping file at the end.

The ED also mentions using `--verbose` to get more details – I'm not sure if there is any other info we can helpfully provide thought?

pkg/upload/unity-ios.go
```diff
@@ -0,0 +1,156 @@
...
+
+		logger.Info(fmt.Sprintf("Found %d dSYM files in: %s", len(dsyms), possibleDsymPath))
+
+		if globalOptions.Upload.UnityIos.VersionName == "" || globalOptions.Upload.UnityIos.BundleVersion == "" || globalOptions.Upload.UnityIos.ApplicationId == "" {
+			plistPath := string(unityOptions.DsymShared.Plist)
+			if plistPath == "" {
+				plistPath = filepath.Join(dsyms[0].Location, "..", "..", "Info.plist")
+			}
+
+			logger.Debug(fmt.Sprintf("Using plist path: %s", plistPath))
```
tomlongridge, line 85: This implies it worked - how about moving the log message inside the `if plistData != nil` check: logged that we _have_read it or else that no plist was found (info only, not an error)

pkg/options/unity.go
```diff
@@ -0,0 +1,35 @@
+package options
+
+import "github.com/bugsnag/bugsnag-cli/pkg/utils"
+
+// UnityLineMapping is used to specify options for uploading a Unity IL2CPP mapping file.
+type UnityLineMapping struct {
+	NoUploadIl2cppMappingFile bool       `name:"no-upload-il2cpp-mapping-file" help:"Do not upload the IL2CPP mapping file (LineNumberMappings.json)"`
```
tomlongridge, line 7: Sorry for late change, but I notice that we don't usually say "path" or "file" in equivalent options, so how about `il2cpp-line-mappings` and `no-upload-il2cpp-line-mappings` as the option names?

pkg/upload/unity-android.go
```diff
@@ -98,16 +106,34 @@ func ProcessUnityAndroid(globalOptions options.CLI, endpoint string, logger log.
...
 				}
+				if filepath.Base(file) == "libil2cpp.so" && !unityOptions.UnityShared.NoUploadIl2cppMappingFile {
+					_, err := elf.GetBuildId(file)
+					if err != nil {
+						return fmt.Errorf("failed to get build ID from %s: %w", file, err)
+					}
+				}
 				symbolFileList = append(symbolFileList, file)
 			}
 		}
```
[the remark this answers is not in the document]
joshedney: Yes, as they have already been processed as part of the Unity build.

tomlongridge (approved):

PR #213, opened 2025-06-23
Add support for Unity iOS and Unity iOS line mappings

joshedney:
## Goal

- Refactor the Xcode build function to allow the Unity iOS command to use parts of it to find the dsyms.
- Add functionality to find the line mappings for an iOS build.
- Add the `unity-ios` command for finding dsyms and line mapping files as part of an Xcode build phase and standalone.
- Add tests to cover the new Unity iOS command.

## Testing

Covered by CI

### DIFF
--- a/pkg/unity/find-line-mappings.go
+++ b/pkg/unity/find-line-mappings.go
@@ -48,3 +48,42 @@ func GetAndroidLineMapping(path string, buildDir string) (string, error) {
 
 	return "", fmt.Errorf("Unable to fine line mapping file in your project: %s ", buildDir)
 }
+
+// GetiOSLineMapping locates the LineNumberMappings.json file for iOS builds.
+//
+// This function attempts to resolve the path to the IL2CPP line number mapping file,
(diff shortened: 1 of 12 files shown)


pkg/unity/find-line-mappings.go
```diff
@@ -48,3 +48,42 @@ func GetAndroidLineMapping(path string, buildDir string) (string, error) {
...
+// Parameters:
+//
+//	path        - an optional explicit path to the mapping file. If provided, it is returned directly.
+//	projectRoot - the root path of the Unity project, used to search for backup artifacts.
+//
+// Returns:
+//
+//	mappingPath - the resolved path to LineNumberMappings.json.
+//	error       - non-nil if the file cannot be found or the backup folder is missing.
+func GetiOSLineMapping(path string, projectRoot string) (string, error) {
```
tomlongridge, line 71: I'm not sure this is the correct casing – `GetIOSLineMapping`?

joshedney: You are correct, everywhere else we have it as `Ios` if its in the middle of a function name

PR #216, opened 2025-07-02
Ensure `create-build` metadata is passed correctly 

joshedney:
## Goal

Ensure that we correctly pass `metadata` to the CLI from the shell and the JS wrapper. 

## Testing

Unit tests and Integration tests added to cover missing areas.

Covered by CI

### DIFF
--- a/js/src/bugsnag-cli-wrapper.ts
+++ b/js/src/bugsnag-cli-wrapper.ts
@@ -25,7 +25,7 @@ class BugsnagCLI {
                     if (typeof value === 'boolean' && value === true) {
                         return [`--${kebabKey}`]
                     } else if (typeof value !== 'boolean') {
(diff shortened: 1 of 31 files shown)


js/src/bugsnag-cli-wrapper.ts
```diff
@@ -25,7 +25,7 @@ class BugsnagCLI {
                     if (typeof value === 'boolean' && value === true) {
                         return [`--${kebabKey}`]
                     } else if (typeof value !== 'boolean') {
-                        return [`--${kebabKey}`, String(value)]
+                        return [`--${kebabKey}=${String(value)}`]
```
tomlongridge, line 28: Can we get a test in to cover this? ... and any other more complex cases you can think of for metadata fields?

PR #225, opened 2025-08-19
Add functionality for Linux (C/C++) symbol files

joshedney:
## Goal

Validate and upload C/C++ symbol files for Linux.

Automatic symbol detection

## Testing

Tests to be added.

### DIFF
--- a/pkg/options/linux.go
+++ b/pkg/options/linux.go
@@ -1,0 +1,13 @@
+package options
+
+import "github.com/bugsnag/bugsnag-cli/pkg/utils"
+
+type LinuxOptions struct {
+	Path          utils.Paths `arg:"" name:"path" help:"The path to the directory or file to upload" type:"path" default:"."`
+	ApplicationId string      `help:"A unique application ID, usually the package name, of the application"`
+	ProjectRoot   string      `help:"The path to strip from the beginning of source file names referenced in stacktraces on the BugSnag dashboard" type:"path"`
+	Variant       string      `help:"The build type/flavor (e.g. Debug, Release) used to disambiguate the between built files when searching the project directory" default:"release"`
+	VersionCode   string      `help:"The version code of this build of the application"`
+	VersionName   string      `help:"The version of the application"`
+	BuildFolder   utils.Path  `help:"The path to the build folder" type:"path"`
+}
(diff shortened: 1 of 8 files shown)


pkg/options/linux.go
```diff
@@ -0,0 +1,12 @@
+package options
+
+import "github.com/bugsnag/bugsnag-cli/pkg/utils"
+
+type LinuxOptions struct {
+	Path          utils.Paths `arg:"" name:"path" help:"The path to the directory or file to upload" type:"path" default:"."`
+	ApplicationId string      `help:"A unique application ID, usually the package name, of the application"`
+	ProjectRoot   string      `help:"The path to strip from the beginning of source file names referenced in stacktraces on the BugSnag dashboard" type:"path" default:"."`
+	VersionCode   string      `help:"The version code of this build of the application"`
```
tomlongridge, line 9: Although it's in the API, we can effectively leave it unused as it's an Android concept. So let's delete this option from the CLI.

pkg/upload/linux.go
```diff
@@ -0,0 +1,129 @@
...
+			if err != nil {
+				return fmt.Errorf("building file list from %q: %w", path, err)
+			}
+			logger.Debug(fmt.Sprintf("Found %d files in directory %s", len(fileList), path))
+		} else {
+			fileList = append(fileList, path)
+		}
+
+		// Filter for valid ELF symbol files
+		for _, file := range fileList {
```
tomlongridge, line 99: Do we only want to do this for `.so` or `.so.debug` files? It's potentially a lot of files to open and read otherwise?

pkg/upload/linux.go
```diff
@@ -0,0 +1,129 @@
...
+//   - error: non-nil if scanning, build ID resolution, or upload fails.
+func ProcessLinux(opts options.CLI, logger log.Logger) error {
+	linuxOpts := opts.Upload.Linux
+
+	var fileList []string
+	var soFileList []string
+	var err error
+
+	for _, path := range linuxOpts.Path {
+		logger.Info(fmt.Sprintf("Scanning path: %s", path))
```
tomlongridge, line 85: This log message doesn't make much sense if you've provided a single file path. How about moving it inside the subsequent if-else to log whether you're scanning a directory or using a single file?

pkg/utils/capitalisation.go
```diff
@@ -0,0 +1,14 @@
+package utils
+
+import (
+	"strings"
+	"unicode/utf8"
+)
+
+func Capitalize(s string) string {
```
tomlongridge, line 8: No longer used?

tomlongridge (approved):

PR #241, opened 2025-10-21
Generic React Native Sourcemaps Command

joshedney:
## Goal

Introduce a generic React Native Sourcemaps upload command that decouples the upload process from automatic discovery.
This command uses configurable escape hatches to supply the necessary metadata for uploading sourcemaps to BugSnag, providing greater flexibility and control.

Existing React Native Android and iOS commands have been refactored to leverage this new generic command for their API uploads.

## Testing

Added new test coverage for the generic Sourcemaps command to validate configuration handling, API integration, and compatibility with platform-specific commands.

### DIFF
--- a/.github/workflows/downstream_updates.yml
+++ b/.github/workflows/downstream_updates.yml
@@ -29,7 +29,7 @@ jobs:
           sudo apt-get install libcurl4-openssl-dev net-tools
 
       - run: >
-          curl -X POST https://api.github.com/repos/bugsnag/bugsnag-unity/dispatches
+          curl -X POST https://api.github.com/repos/${{ matrix.downstream_repo }}/dispatches
           -H 'Content-Type: application/json'
           -H "Authorization: Bearer ${{ secrets.DEP_UPDATER_BEARER_TOKEN }}"
           -d '{


pkg/upload/react-native-sourcemaps.go
```diff
@@ -0,0 +1,111 @@
...
+			uploadOpts["appVersionCode"] = reactNativeOpts.VersionCode
+		} else if reactNativeOpts.BundleVersion != "" {
+			logger.Debug(fmt.Sprintf("Using iOS bundle version: %s", reactNativeOpts.BundleVersion))
+			uploadOpts["appBundleVersion"] = reactNativeOpts.BundleVersion
+		}
+	}
+
+	// Development mode flag
+	if reactNativeOpts.Dev {
+		logger.Debug("Development build detected — marking upload as dev=true")
```
tomlongridge: I might be wrong, but I don't think we're actually detecting this?

```suggestion
		logger.Debug("Development build — marking upload as dev=true")
```

pkg/upload/react-native-sourcemaps.go
```diff
@@ -0,0 +1,111 @@
...
+	if reactNativeOpts.ProjectRoot == "" {
+		uploadOpts["projectRoot"] = string(reactNativeOpts.Path)
+		logger.Debug(fmt.Sprintf("Project root not provided — using inferred path: %s", uploadOpts["projectRoot"]))
+	} else {
+		uploadOpts["projectRoot"] = reactNativeOpts.ProjectRoot
+		logger.Debug(fmt.Sprintf("Using specified project root: %s", reactNativeOpts.ProjectRoot))
+	}
+
+	// Platform and overwrite flag
+	uploadOpts["platform"] = reactNativeOpts.Platform
```
tomlongridge: Is `Platform` a required field or could it be empty? It feels like we should validate this somewhere higher up, checking that it's one of our 3 supported ones?

joshedney: This validation has been added to the `utils` package

PR #249, opened 2025-12-04
add fallback path for androidmanifest.xml android proguard

joshedney:
## Goal

Update how we check for the AndroidManifest.xml file for the `upload android-proguard` command by checking a fallback path if it is not present in the  default location.

## Testing
Covered by CI

### DIFF
--- a/pkg/android/find-android-manifest.go
+++ b/pkg/android/find-android-manifest.go
@@ -1,6 +1,7 @@
 package android
 
 import (
+	"fmt"
 	"path/filepath"
 
 	"github.com/bugsnag/bugsnag-cli/pkg/utils"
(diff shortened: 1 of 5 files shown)


pkg/android/find-android-manifest.go
```diff
@@ -0,0 +1,24 @@
...
+// This function iterates through a list of candidate file paths and returns the
+// first one that exists on disk. This is useful when Android build outputs may
+// place the manifest in multiple variant-specific directories.
+//
+// Parameters:
+//   - paths: A slice of possible manifest file locations.
+//
+// Returns:
+//   - string: The first existing manifest path, or an empty string if none are found.
+func FindAndroidManifest(paths []string) string {
```
tomlongridge: Could this function rather take the common merged manifest directory and variant name and look in the two known locations?

```
<merged_manifest_path>/<variant>/AndroidManifest.xml
<merged_manifest_path>/<variant>/process<variant>Manfiest/AndroidManifest.xml
```

joshedney: I've updated the function to accept the merged_manifest_path and variant. It then builds and checks the paths within the function.

pkg/android/find-android-manifest.go
```diff
@@ -0,0 +1,38 @@
...
+
+// FindAndroidManifest locates and returns the first existing AndroidManifest.xml
+// for a given build directory and variant.
+//
+// This function constructs and checks multiple potential manifest locations that
+// vary based on Android build outputs. It returns the first manifest file found
+// on disk.
+//
+// Parameters:
+//   - path: The Android merged_manifests directory.
```
tomlongridge: nit: Can we just tweak it slightly so it's the build directory that's passed in and you add "intermediates/merged_manifests" in the function?

Negligible functional benefit, but logically it's nice that `FindAndroidManifest` will have all the knowledge of where to try and locate the file.

I'd also suggest renaming the parameter as `appBuildPath` for clarity.

pkg/android/find-android-manifest.go
```diff
@@ -0,0 +1,38 @@
...
+//   - variant: The build variant (e.g., "debug", "release") whose manifest paths
+//     should be searched.
+//
+// Returns:
+//   - string: The resolved manifest path if found, otherwise an empty string.
+func FindAndroidManifest(path string, variant string) string {
```
tomlongridge: Can this be used elsewhere where we're locating a manifest file?

https://github.com/search?q=repo%3Abugsnag%2Fbugsnag-cli%20%22merged_manifests%22&type=code

joshedney: I've updated other locations where we're locating the Android manifest so that they now use this finder function.

tomlongridge (approved):

PR #250, opened 2025-12-09
Add Node.JS integration tests

joshedney:
## Goal

Add Node.JS integration tests with a minimal fixture. 

## Testing

Covered by CI

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -155,6 +155,19 @@ steps:
         upload:
           - maze_output/**/*
 
+  - label: "Node.js Integration Tests"
+    depends_on: build
+    commands:
+      - chmod +x bin/arm64-macos-bugsnag-cli
+      - bundle install
+      - bundle exec maze-runner --port=$((MAZE_RUNNER_PORT)) features/node
+    plugins:
+      artifacts#v1.5.0:
+        download:
+          - bin/arm64-macos-bugsnag-cli
+        upload:
+          - maze_output/**/*
+
   - group: ":react: React Native"
     steps:
       - label: ":large_blue_circle: :large_blue_circle: :large_blue_circle: REACT NATIVE ANDROID STEPS :large_blue_circle: :large_blue_circle: :large_blue_circle:"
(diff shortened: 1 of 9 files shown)


features/node/node.feature
```diff
@@ -0,0 +1,35 @@
+Feature: Node.js Integration Tests
+  @CleanAndBuildNodeJs
```
twometresteve: It feels like this would be suited better to a `Background` step calls `I clean and build Node JS`.  We've tended to use Cucumber tags more to decide if a scenario should be run or not, rather than performing common steps.

features/steps/steps.rb
```diff
@@ -188,6 +188,10 @@ def get_version_number(file_path)
   Maze.check.not_include(run_output, "Error 1")
 end
 
+And(/^I wait for the Node\.js sourcemaps to generate$/) do
```
twometresteve: The name seems misleading as it's not waiting for anything, just checking if it is there.  You could use a `Maze::Wait` if you really want a wait, but perhaps it's not necessary.

features/steps/steps.rb
```diff
@@ -463,3 +467,15 @@ def clean_and_build(scheme, project_path, build_target)
...
+
+Before('@CleanAndBuildNodeJs') do
+  @fixture_dir = "#{base_dir}/features/node/fixtures/"
+  Dir.chdir(@fixture_dir)
+  @output =`npm run clean && npm i && node build.js`
+  Dir.chdir(base_dir)
+
+  steps %(
+    And I wait for the Node.js sourcemaps to generate
+  )
```
twometresteve: Do we really need to call out to another step here?  Can we just put that code inline here?

.buildkite/pipeline.yml
```diff
@@ -155,6 +155,19 @@ steps:
...
+    commands:
+      - chmod +x bin/arm64-macos-bugsnag-cli
+      - bundle install
+      - bundle exec maze-runner --port=$((MAZE_RUNNER_PORT)) features/node
+    plugins:
+      artifacts#v1.5.0:
+        download:
+          - bin/arm64-macos-bugsnag-cli
+        upload:
+          - maze_output/**/*
```
twometresteve: Not a big deal for just 2 scenarios, but this could pay off longer term or in other situations that just copy/paste this step as an example:
```suggestion
          - maze_output/maze_output.zip
          - maze_output/failed/*
```

twometresteve (approved):

PR #259, opened 2026-01-30
Handle postinstall binary download error

trvrnrth:
## Goal

A failure to fetch from GitHub should fail the postinstall rather than silently leaving the stub bin in place.

## Testing

Happy case covered by CI.


trvrnrth:
@joshedney This needs work, but it was something I stumbled across when testing in an environment with flakey connectivity.

PR #269, opened 2026-03-06
Add `--exclude` option to filter files during uploads

joshedney:
## Goal

Add support for excluding files from uploads based on pattern matching. This allows users to filter out unwanted files (e.g., development source maps, temporary files, or specific directories) during upload operations without having to manually organize their file structure.

## Design

The `--exclude` option was implemented at three levels:

1. **CLI Option Definition** - Added `Exclude []string` field to the `Upload` struct in `pkg/options/options.go`, making it available for all upload commands
2. **Core Matching Logic** - Implemented `IsFileExcluded()` function in `pkg/utils/files.go` that supports multiple types of patterns:
   - **Wildcard matching** using `doublestar.Match()` for file extensions and patterns (e.g., `*.map`, `*.log`)
   - **Recursive globbing** using `**` for deep directory matching (e.g., `node_modules/**`, `**/*.map`, `**/dist/**`)
   - **Exact filename matching** for specific files (e.g., `test.map`)
3. **Upload Integration** - Added exclusion check in `ProcessFileRequest()` in `pkg/server/request.go` before files are uploaded

The implementation uses the `github.com/bmatcuk/doublestar/v4` library for advanced glob pattern matching, including `**` support for recursive directory traversal. This provides powerful and flexible filtering capabilities similar to `.gitignore` patterns, while maintaining compatibility with various file path formats across platforms.

### Pattern Examples

- `*.map` - Exclude all files ending in `.map`
- `node_modules/**` - Exclude all files under `node_modules/` directory
- `**/node_modules/**` - Exclude `node_modules/` at any directory level
- `**/*.test.js` - Exclude all `.test.js` files anywhere in the tree
- `src/**/*.map` - Exclude all `.map` files within the `src/` directory tree
- `**/temp/**` - Exclude all files in any `temp/` directory

## Changeset

- Added `Exclude []string` field to `Upload` struct in `pkg/options/options.go`
- Implemented `IsFileExcluded()` function in `pkg/utils/files.go` with `doublestar` library
- Added `github.com/bmatcuk/doublestar/v4` dependency to `go.mod`
- Added exclusion check in `ProcessFileRequest()` in `pkg/server/request.go`
- Added comprehensive unit tests in `test/utils/files_test.go`:
  - 15 test cases covering wildcards, exact matches, substrings, `**` globbing patterns, and edge cases
- Added integration test in `test/upload/js_test.go`
- Added BDD feature tests in `features/cli/exclude-option.feature` with 5 scenarios

## Testing

Covered by CI


pkg/utils/files.go
```diff
@@ -51,6 +51,37 @@ func IsDir(path string) bool {
...
+		// Try matching against the base name
+		matched, err = filepath.Match(pattern, filepath.Base(filePath))
+		if err == nil && matched {
+			return true
+		}
+
+		// Check if the pattern contains path separators and matches as substring
+		if strings.Contains(filePath, pattern) {
+			return true
+		}
```
tomlongridge: I wonder if this might be too eager to match. It might be more expected for users to need to give a full path match for cases like `node_modules/**` and avoid unexpected excluding.

pkg/options/options.go
```diff
@@ -117,9 +117,10 @@ type Breakpad struct {
 
 type Upload struct {
 	// shared options
-	Retries          int    `help:"The number of retry attempts before failing an upload request" default:"0"`
-	Timeout          int    `help:"The number of seconds to wait before failing an upload request" default:"300"`
-	UploadAPIRootUrl string `help:"The upload server hostname, optionally containing port number"`
+	Retries          int      `help:"The number of retry attempts before failing an upload request" default:"0"`
+	Timeout          int      `help:"The number of seconds to wait before failing an upload request" default:"300"`
+	UploadAPIRootUrl string   `help:"The upload server hostname, optionally containing port number"`
+	Exclude          []string `help:"Exclude files matching these patterns (supports wildcards like *.map, path/to/*, etc.)"`
```
tomlongridge: How consistent are the file paths when they are matched? Is it possible to say in the usage whether the exclude patterns should be relative to where the command is run?

pkg/options/options.go
```diff
@@ -117,10 +117,10 @@ type Breakpad struct {
...
 type Upload struct {
 	// shared options
-	Retries          int    `help:"The number of retry attempts before failing an upload request" default:"0"`
-	Timeout          int    `help:"The number of seconds to wait before failing an upload request" default:"300"`
-	UploadAPIRootUrl string `help:"The upload server hostname, optionally containing port number"`
-
+	Retries          int      `help:"The number of retry attempts before failing an upload request" default:"0"`
+	Timeout          int      `help:"The number of seconds to wait before failing an upload request" default:"300"`
+	UploadAPIRootUrl string   `help:"The upload server hostname, optionally containing port number"`
+	Exclude          []string `help:"Exclude files matching these patterns. Supports wildcards (*.map), recursive globs (node_modules/**, **/*.test.js) and exact filenames (file.js.map)."`
```
tomlongridge:
```suggestion
	Exclude          []string `help:"Exclude files matching these patterns. Supports wildcards (*.map), recursive globs (node_modules/**, **/*.test.js) and exact filenames (file.js.map). Non-absolute path patterns are relative to the current directory."`
```

pkg/android/process-uploads.go
```diff
@@ -92,7 +92,7 @@ func UploadAndroidNdk(
 			"/ndk-symbol",
 			params,
 			fileField,
-			filepath.Base(originalFile),
```
tomlongridge, line 95: Does this change not have a functional effect? Or does it fix an existing limitation for nested files?

tomlongridge (commented):
Can you also rebase against next?

pkg/android/process-uploads.go
joshedney: It was a duplicate of what we're already doing in the `ProcessFileRequest`. Removing it from here means that we're able to keep the exclude check at the point of upload rather than moving it higher up.

tomlongridge (approved):

PR #270, opened 2026-03-06
Enforce HTTP/1.1

joshedney:
## Goal

Configure the Bugsnag CLI to use HTTP/1.1 instead of HTTP/2 for all API requests to the upload and build endpoints. This ensures compatibility with network infrastructure that may have issues with HTTP/2 or require HTTP/1.1 for monitoring, proxying, or debugging.

## Design

The HTTP client is configured to use HTTP/1.1 only by leveraging Go 1.24+'s `http.Protocols` API. This approach:

- Explicitly sets which HTTP protocols to use via `SetHTTP1(true)` and `SetHTTP2(false)`
- Forces all connections to use HTTP/1.1
- Uses the modern, future-proof API for protocol selection
- Maintains backward compatibility with existing functionality
- Requires minimal code changes (only in `pkg/server/request.go`)

The `http.Protocols` type provides a clear, type-safe way to configure supported protocols, making the code more maintainable and explicit about its behavior compared to older workarounds.

## Changeset

- **Updated `go.mod`**: 
  - Bumped Go version from 1.23.1 to 1.25.4 to support `http.Protocols` API

- **Modified `pkg/server/request.go`**: 
  - Updated `sendRequest()` function to create a custom `http.Transport` with `Protocols` configured for HTTP/1.1 only
  - Added inline comment explaining the HTTP/1.1 configuration

## Testing
Covered by CI


pkg/server/request.go
```diff
@@ -257,8 +258,14 @@ func processRequest(request *http.Request, timeout int, retryCount int, logger l
 // Returns:
 //   - error: An error if any step of the request processing fails. Nil if the process is successful.
 func sendRequest(request *http.Request, timeout int, logger log.Logger) error {
+	// Configure transport to use HTTP/1.1 only
+	transport := &http.Transport{
```
tomlongridge, line 265: The [golang docs](https://pkg.go.dev/net/http#Transport) suggest using `Transport.Protocols` now. That feels like it would be more future-proof for as/when the default changes.

tomlongridge (approved):

PR #275, opened 2026-03-25
fix: rebuild HTTP requests for each retry attempt

joshedney:
## Goal

Fix HTTP request retry failures that occurred when upload requests needed to be retried. When retries were enabled (`--retries > 0`), subsequent retry attempts would fail with the error: `error sending request: Post "https://upload.bugsnag.com:443/ndk-symbol": http: ContentLength=1278529 with Body length 0`

This prevented users from successfully uploading symbol files when transient network issues or duplicate file conflicts occurred.

## Design

The root cause was that HTTP request bodies (io.Reader) can only be read once. When a request failed and needed to be retried, the body stream had already been consumed by the first attempt, leaving an empty body for subsequent retries.

The fix implements a **builder pattern** approach that reconstructs a fresh HTTP request for each retry attempt:

1. **Introduced requestBuilder type**: A function type that constructs a fresh HTTP request on demand
2. **Refactored processRequest()**: Changed to accept a `requestBuilder` function instead of a pre-built request, calling it before each attempt
3. **Updated callers**: Modified `ProcessFileRequest()` and `ProcessBuildRequest()` to pass builder functions that construct fresh requests

This approach entirely avoids request state pollution by creating a new request object for each attempt, rather than trying to reuse and reset an existing request. This is cleaner and more robust than manually managing request body state, as it ensures headers, body, and all internal state are fresh for each retry.

## Changeset

- **`pkg/server/request.go`**:
  - Added `requestBuilder` function type for constructing fresh HTTP requests
  - Modified `buildFileRequest()`: Simplified to build standard requests without GetBody logic
  - Modified `processRequest()`: Now accepts a `requestBuilder` and builds a fresh request for each retry attempt
  - Modified `ProcessFileRequest()`: Passes a closure that calls `buildFileRequest()` for each attempt
  - Modified `ProcessBuildRequest()`: Passes a closure that creates a fresh request with the payload for each attempt

## Testing

**Manual Testing:**
- Tested with the command: `./bin/arm64-macos-bugsnag-cli upload unity-android platforms-examples/Unity/UnityExample-1.0-v1-IL2CPP.symbols.zip --api-key=<key> --retries=3 --verbose`
- Verified that retries now successfully re-send the request body instead of failing with "Body length 0" errors
- Confirmed that the fix compiles successfully with `make build`

**Automated Testing:**
Covered by CI


pkg/server/request.go
```diff
@@ -237,6 +254,7 @@ func processRequest(request *http.Request, timeout int, retryCount int, logger l
 		}
 
 		logger.Warn("Request Failed, Retrying...")
+		logger.Warn(err.Error())
```
tomlongridge: I think it might be better to wrap the error so it's clearer when looking at the logs that it's a retry.

e.g.

```
BugSnag API request attempt 1 failed:
<error message>
Retrying...
```

tomlongridge (approved):

PR #286, opened 2026-05-25
plat-16005 andorid manifest fix

SB-jigneshR:
## Fixed
- Correctly handle capitalization in fallback AndroidManifest.xml search path (e.g., processStagingReleaseManifest) to support case-sensitive filesystems like Linux/Alpine. Previously, the search used incorrect capitalization, causing manifest detection to fail on some systems.


SB-sachinp (approved):
LGTM

